PR #544, opened 2014-04-09
Added inclusion of aggregations to pagination

r1pp3rj4ck:
Since ElasticSearch 1.0 there is a support for aggregation which are pretty much new generation facets. Ruflin's Elastica supports them, so I added it to the pagination the same way as the facets are.


merk:
3.1-dev contains these changes

Issue #557, opened 2014-04-20
[RFC] New 'API' for using the bundle


merk:
## New Resultset Api

We've had a few issues previously, like not being able to access highlights, or other information about something that Elastica processes and exposes through its ResultSet. I think it makes more sense for us to harmonise with this method of accessing results, and theres no reason our Transformations cant be shoehorned into a ResultSet/Result combo.

It means that we'd deprecate and remove the Finder services, and move back to just exposing the index and type services that return ResultSets that contain lazy (or not so lazy) transformations based on FOSElasticaBundle's configuration.

I'd like users feedback on this - it will be a BC break and it has to be decided if we make this change for 3.0, or finish off the remaining issues with 3.0, release it and then start working on this feature for 4.0.

I'd like to see this land sooner rather than later - we might be able to make this BC with some trickery for 3.1, but the cycle between 3.0, 3.1 and 4.0 in this case would only be a few weeks to a few months.
## Example

``` php
<?php

/**
 * We override the Type class used for creating or retrieving new types, so that we
 * can return our ResultSet instead of Elastica's.
 *
 * @var \FOS\ElasticaBundle\Elastica\TransformableType
 */
$type = $this->container->get('fos_elastica.type.customer');

/**
 * Our resultset extends Elasticas..
 *
 * @var \FOS\ElasticaBundle\Elastica\TransformableResultSet
 */
$resultSet = $type->search(/* ... */);

foreach ($resultSet as $result) {
    /**
     * The Result object is the same as the Elastica result, except it has an
     * additional property, $transformed, that will (potentially lazily) load the
     * Transformed document as the current transformers work.
     *
     * @var \FOS\ElasticaBundle\Elastica\TransformableResult $result
     */

    // Retrieve highlight information from the Elastica interface
    $highlights = $result->getHighlights();

    // Lazily load the Customer entity from the persistence provider - could
    // be configured to load all ResultSet transformations on first access, or
    // lazily for each Result in the ResultSet
    $customer = $result->getTransformed();
}

// Retrieve an array of transformed results directly from the ResultSet
$customers = $resultSet->getTransformed();
```

Tobion:
Closing as no progess has been made in a while.

PR #559, opened 2014-04-24
[POC][4.0] Internal refactoring + new api for transformations

merk:
Relating to #557, I've spent some time just mucking around with code to see what it might look like.

There is a commit in here about deprecating some top level classes that will make it into 3.0 but I'd like to see if people have any comments about what I've done here. The second commit has all the changes relating to the internal refactoring/new api.


Tobion:
Closing as it conflicts and seems to be stale.

PR #633, opened 2014-06-25
Added query expand/shrink with pretty print and code highlighting for easy query debugging

aurimasniekis:


XWB:
This PR is outdated, closing.

PR #660, opened 2014-07-14
Add aggregations with paginator

rudrud:
only facets were implemented, why not aggregations ?


sroze:
@merk Note that there's no tests on the facets part, what kind of test(s) do you want ?

merk:
This is merged into 3.1-dev

PR #664, opened 2014-07-16
Update ModelToElasticaAutoTransformer

agounaris:
It would be good if empty values are not indexed for the target document.


merk:
An empty value is valid.

PR #665, opened 2014-07-17
Issue: Parent is missing in the fields list that causes RoutingMissingEx...

### LINKED ISSUE #774: [Regression] The configuration refactoring broke the parent child mapping
After https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/629, the transformer does not receive the ``_parent`` configuration anymore, and so does not set the parent when transforming the entity to a document, thus breaking the indexation

tumbochka:
...ception on flushing

Having config:

```
                extra_field:
                    mappings:
                        text: ~
                        title: ~
                    _parent: { type: "user", property: "owner", identifier: "id" }
                    persistence:
                        driver: orm
                        model: Foo\CoreBundle\Entity\ExtraInfo
                        provider: ~
                        listener: ~
                        finder: ~
```

When building the project getting an exception on fixture loading when flushing ExtraInfo: 
```                                                                                      
  RoutingMissingException[routing is required for [foo_dev]/[extra_field]/[1]] 
```

The  problem was that ObjectPersister got only the entity fields without the _parent field.
The reason was that the service in the ProjectContainer was generated incorrectly.

I'm not sure it's the best way to fix, please correct me if I'm wrong.

Closes #774

### DIFF
--- a/DependencyInjection/FOSElasticaExtension.php
+++ b/DependencyInjection/FOSElasticaExtension.php
@@ -391,7 +391,13 @@ class FOSElasticaExtension extends Extension
             $arguments[] = array(new Reference($callbackId), 'serialize');
         } else {
             $abstractId = 'fos_elastica.object_persister';
-            $arguments[] = $this->indexConfigs[$indexName]['types'][$typeName]['mapping']['properties'];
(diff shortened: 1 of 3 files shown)


fancyweb:
:+1: We seriously need this fix.

stof:
I merged this in the 3.0.x branch

PR #681, opened 2014-08-05
Ensure Doctrine objects with collection (ReferenceMany) changes are updated in Elasticsearch

zarathustra323:
This PR adds support to the Doctrine Listener to ensure that when a Collection (e.g. ReferenceMany) is modified on an Object (Entity/Document) it is updated in Elasticsearch.

Use case:
- If Object A has a ReferenceMany that is modified (added to or removed from) it currently doesn't schedule an Elasticsearch update for Object A.
- This is because the UnitOfWork does not fire a postUpdate event on changes to collections, but rather must be processed postFlush as a Collection Update or Removal.

The update and insertion logic for scheduling an Object was moved into their own methods for reusability. These methods also now key the scheduled Object by its spl_object_hash. This was done to ensure that if an Object has multiple collection changes it is only sent to the ObjectPersister once.

Similar to @solocommand's getDoctrineObject, the getObjectManager, getUnitOfWork, and getCollectionChanges methods ensure that the appropriate getter is used to retrieve the corresponding service. These utilize method_exists checks since the interfaces don't define the methods (though they exist in both the ORM and ODM versions). Please view in-line docblocks for more information.

### DIFF
--- a/Doctrine/Listener.php
+++ b/Doctrine/Listener.php
@@ -117,27 +117,70 @@ class Listener implements EventSubscriber
         throw new \RuntimeException('Unable to retrieve object from EventArgs.');
     }
 
+    /**
+     * Provides unified method for retrieving a doctrine object manager from an EventArgs instance
+     *
+     * @param   EventArgs           $eventArgs
+     * @return  ObjectManager       An instance implementing ObjectManager
+     * @throws  \RuntimeException   if no valid getter is found.
+     */
+    private function getObjectManager(EventArgs $eventArgs)
+    {
+        if (method_exists($eventArgs, 'getObjectManager')) {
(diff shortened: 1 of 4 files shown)


merk:
@zarathustra323 I've just been going over this PR (sorry, its been a while) and I cant seem to find a situation where a collection can be marked for update/delete in the UnitOfWork where the owning entity doesnt also get checked for changes. This means that most of the PR doesnt actually cause any different behaviour.

Do you happen to remember anything specific about how you arrived at using the UoW's tracking of collections?

XWB:
Closing this PR as it is completely outdated.

aLkRicha:
will this be done?

PR #725, opened 2014-10-08
Fix mongodb populate falling down performances for big collections

Gnucki:
Using skip to retrieve slices of data is not performant when the amount of documents skipped is important. It decreases the performances of the command `fos:elastica:populate` for each retrieved slice until it crashes (30s timeout).

http://docs.mongodb.org/manual/reference/method/cursor.skip/

To fix this problem, a solution is to retrieve a slice from the last object of the previous slice. Mongo can then use an index to find the slice instead of browsing all the collection.


merk:
Replaced by #816

PR #728, opened 2014-10-16
Add missing KnpPaginator example in the usage documentation

ugomeda:
The example for the Knp Paginator was missing in the usage documentation.


phreaknerd:
@merk Where can this docs be found?

PR #729, opened 2014-10-18
Avoid Doctrine\Listener::getSubscribedEvents() call 

korotovsky:
... on each page where doctrine is active.

The problem:

Classes FOSElasticaBundle shouldn't be loaded (providers, indexes, http client) until first call.

If there's at least one `Doctrine\Listener` registered in the application, then the following code for `\Symfony\Bridge\Doctrine\ContainerAwareEventManagerDoctrine` will be generated in `app/cache/dev/appDevDebugProjectContainer.php`:

    $c = new \Symfony\Bridge\Doctrine\ContainerAwareEventManager($this);
    $c->addEventSubscriber($this->get('fos_elastica.listener.index.type'));

Inside `addEventSubscriber()` non-static `Doctrine\Listener::getSubscribedEvents()` is called for each subscriber.

As far as we use proxy-services, after `Doctrine\Listener::getSubscribedEvents()` is called, dependencies for `Doctrine\Listener` start resolving.


merk:
Thanks. I've merged this into 3.1-dev and adjusted the mongodb event listener configuration because the changes made here would have broken mongodb users.

PR #731, opened 2014-10-30
Improve docs

fpondepeyre:
Explain "search" key configuration


stof:
@fpondepeyre ping again

PR #732, opened 2014-11-01
Ability to set round_robin for elastica clients

merk:


merk:
Just a ping on this issue. I've merged it into 3.0.x instead of master.

PR #744, opened 2014-11-21
add index populate events

oleg-andreyev:
In my project we are using bulk indexing, and according to [doc](http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/indices-update-settings.html#bulk) to increase performant for bulk indexing, we need to have ability call custom actions like: `refresh_interval` and `_optimize?max_num_segments=5`

To achieve that we've added `PRE_INDEX_POPULATE`, `PRE_TYPE_POPULATE`, `POST_TYPE_POPULATE`, `POST_INDEX_POPULATE`, `INDEX_RESET` events.

### DIFF
--- a/Command/PopulateCommand.php
+++ b/Command/PopulateCommand.php
@@ -2,15 +2,16 @@
 
 namespace FOS\ElasticaBundle\Command;
 
+use FOS\ElasticaBundle\Event\PopulateEvent;
 use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
 use Symfony\Component\Console\Helper\DialogHelper;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use FOS\ElasticaBundle\IndexManager;
+use FOS\ElasticaBundle\Index\IndexManager;
 use FOS\ElasticaBundle\Provider\ProviderRegistry;
-use FOS\ElasticaBundle\Resetter;
 use FOS\ElasticaBundle\Provider\ProviderInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 /**
  * Populate the search index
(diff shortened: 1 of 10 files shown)


merk:
@oleg-andreyev ping

EventListener/PopulateListener.php
```diff
@@ -0,0 +1,60 @@
...
+     * @var Resetter
+     */
+    private $resetter;
+
+    /**
+     * @param Resetter $resetter
+     */
+    public function __construct(Resetter $resetter)
+    {
+
```
oleg-andreyev: done

Command/PopulateCommand.php
```diff
@@ -141,19 +130,48 @@ private function populateIndex(OutputInterface $output, $index, $reset, $options
...
+
+    /**
+     * @param OutputInterface     $output
+     * @param ProviderInterface[] $providers
+     * @param string              $index
+     * @param string              $type
+     * @param boolean             $reset
+     * @param array               $options
+     */
+    private function populate(OutputInterface $output, $providers, $index, $type, $reset, $options)
```
merk: Since providers is always required we can typehint array here

merk:
It looks like you've removed the RESET event entirely, I think it still has a useful purpose but needs to fire inside the Resetter.

oleg-andreyev:
@merk, I've noticed that `ResetterTest` is disabled, is there a ticket for that, I can address it.

merk:
I dont believe there is at the moment, I would appreciate it if you could look into fixing the tests for that class. Just create a new PR

dupuchba:
any news on this one ?

oleg-andreyev:
ping @merk

merk:
@oleg-andreyev I've done a bit more work to get this mergable with current master - Any chance you could work on a cookbook entry to add to the docs so there is some documentation on how to use these events?

A pr for others to review based off this one: #814

oleg-andreyev:
Okay, I'll create new PR for cookbook

Issue #746, opened 2014-11-26
Incorrect Total Hits from Paginator


NinjDS:
The FOS\ElasticaBundle\Paginator\RawPaginatorAdapter::getTotalHits() method is alienating the Elastic Search behaviour.

**Introduction**

Elastic Search is providing a `hits.total` value that is used *to know how many documents match the sent query*. This is an important fact, because it allows ES users to be aware of the total count, regardless of the number of documents they actually got in the response.

This total hits count must be useable.

**The current behaviour**

FOSElasticaBundle takes a decision in place of the developer: change the total hits count if a size was provided in the request. Here is where it happens:

    public function getTotalHits()
    {
        if ( ! isset($this->totalHits)) {
            $this->totalHits = $this->searchable->search($this->query)->getTotalHits();
        }

        return $this->query->hasParam('size')
            ? min($this->totalHits, (integer) $this->query->getParam('size'))
            : $this->totalHits;
    }

The documentation reads "Returns the number of results". True, it does return the number of results. So why not call it "getResultCount"? Or "getNumberOfResults" ? "Total Hits" has already a different meaning in Elastic Search, it's a mistake to do so, in my opinion.

**Expected behaviour**

In order not to introduce any BC break, i see two solutions:

OPTION 1: Add an optional parameter to getTotalHits:

    public function getTotalHits($onlyResults = true)
    {
        if ( ! isset($this->totalHits)) {
            $this->totalHits = $this->searchable->search($this->query)->getTotalHits();
        }

        return $this->query->hasParam('size') && $onlyResults
            ? min($this->totalHits, (integer) $this->query->getParam('size'))
            : $this->totalHits;
    }

OPTION 2: Introduce a new method:

    public function getRealTotalHits()
    {
        if ( ! isset($this->totalHits)) {
            $this->totalHits = $this->searchable->search($this->query)->getTotalHits();
        }

        return $this->totalHits;
    }

Thank you for any review, i'm open to discussions and remarks of course.

PR #748, opened 2014-11-26
Genuine ES behaviour for getTotalHits()

NinjDS:
In response to the issue #746

This minor edit allows RawPaginatorAdapter::getTotalHits() to return the real total count (i.e. all matching documents).

By default, the total is limited to the size query parameter, which doesn't comply with elasticsearch api.

It is useful when you only need to count the number of documents that correspond to a given query, without actually retrieving any document.


iBasit:
any docs on how to use this on `3.2.2`

iBasit:
Why these changes are not showing in 3.2.2 but showing in 4.x?

Issue #287, opened 2013-05-10
fos:elastica:populate doesn't use is_indexable_callback


arnestockmans:
In my configuration I use `is_indexable_callback` several times. This seems to be working (except for issue #284), although when I run the fos:elastica:populate command, this setting gets ignored completely.

merk:
The documentation is up to date. `indexable_callback` is the new parameter name. `is_indexable_callback` is the old name.

https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/DependencyInjection/Configuration.php#L230

PR #755, opened 2014-12-05
Update Documentation for is_indexable_callback

wodka:
as written in #287 the documentation is wrong for the is_indexable_callback.

thx @Merk87 for spotting

### DIFF
--- a/Resources/doc/types.md
+++ b/Resources/doc/types.md
@@ -157,7 +157,7 @@ different kinds of objects if your persistence backend supports these methods,
 but in some cases you might want to run an external service or call a property
 on the object to see if it should be indexed.
 
-A property, `indexable_callback` is provided under the type configuration that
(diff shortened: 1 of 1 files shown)


merk:
Thanks, however this is not correct: indexable_callback is the correct configuration parameter. is_indexable_callback is a deprecated parameter from old versions.

wodka:
Hi, 

then I'm curious why I see the following error on dev-master:

```
Unrecognized option "indexable_callback" under "fos_elastica.indexes.search.types.article.persistence.listener"
```

(running Version c39b86c6c5e229271d718671560ba63d288d0965)

merk:
Because the location should be fos_elastica.indexes.search.types.article.indexable_callback

wodka:
thx, that solved it

PR #760, opened 2014-12-17
Fix PHP notice when using indexes without defined types

michaelperrin:
This fixes a PHP notice that comes up when executing the `fos:elastica:populate` command if one on the indexes in `config.yml` doesn't define any associated type. Example:

```yaml
fos_elastica:
    indexes:
        my_index: ~
        my_index2:
            types:
                # ...
```

I didn't add any unit test for this, as this is just a simple bug.

### DIFF
--- a/Configuration/Source/ContainerSource.php
+++ b/Configuration/Source/ContainerSource.php
@@ -41,13 +41,16 @@ class ContainerSource implements SourceInterface
         $indexes = array();
         foreach ($this->configArray as $config) {
             $types = array();
-            foreach ($config['types'] as $typeConfig) {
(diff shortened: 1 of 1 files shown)


merk:
Merged with aef5940578d63af22f4129c3c186ce22ed4ea66a

stof:
I would rather have fixed the DI extension to ensure that the types key is always set with an empty array

merk:
I plan on deprecating the ability to configure the bundle's indexes and types from the container at some point in the future. To me, this kind of metadata matches the way Doctrine's is defined.

merk:
*should match

PR #768, opened 2015-01-07
Fix ODM listener service arguments

vkartaviy:
It became broken after https://github.com/FriendsOfSymfony/FOSElasticaBundle/commit/d731443aa5ad7a12c133bd1d7b97fa242c546e1e commit


merk:
Thanks

Issue #766, opened 2015-01-04
Migrate populate command to use ProgressBar


merk:
Instead of using our own output, i'd prefer to see a ProgressBar if it is available.

pmartelletti:
@merk something like this?

http://symfony.com/doc/2.3/components/console/helpers/progresshelper.html

If so, I will try to create a PR. :smile:

merk:
The ProgressHelper is deprecated in favour of the ProgressBar, but the ProgressBar is not available in 2.3 or 2.4.

We still support symfony 2.3.

http://symfony.com/doc/current/components/console/helpers/progressbar.html

pmartelletti:
@merk that's why I suggested using ProgressHelper instead of ProgressBar (that looks nicer). Or are you willing to find some external component?

merk:
I dont want to implement a deprecated feature - we're going to have to evaluate 3.0 support in the next few months.

I'm not sure how hard it would be to use the ProgressBar when it is present or just default back to what we display at the moment if not..

pmartelletti:
@merk in order to use the progress bar, I'll probably need to make the Provider method 'countObject' public. Is there a chance I can change that?

If not, progress bar can be used as a throbber instead of a progress bar.

http://symfony.com/doc/current/components/console/helpers/progressbar.html

pmartelletti:
@merk I've made some tests for the ORM provider, and if we create a method like this:

``` php
/**
     * @return int|mixed
     */
    public function getTotalObjects()
    {
        return $this->countObjects($this->createQueryBuilder());
    }
```

we could implement the symfony progress bar, and it's working pretty well. Do you agree with this method (that could be implemented on the AbstractProvider)? Or do you have any ideas for a workaround?

pmartelletti:
@merk created the PR. Let me know about it. :smile:

royraider:
Hello !
I'm in Symfony 2.5.9 and I got this when trying to "populate":
php app/console fos:elastica:populate
> Resetting app                                    
  [LogicException]              
  You must start the progress bar before calling setCurrent().

I fixed this by disabling the ProgressBar. (line 128 of populateCommand)

merk:
Please call the command with --verbose

BourotBenjamin:
Here's what we got


     [LogicException]
     You must start the progress bar before calling setCurrent().

    Exception trace:
[...]

PR #770, opened 2015-01-07
Use ProgressBar in populate commande when available.

pmartelletti:
Work for #766 .

### DIFF
--- a/Command/PopulateCommand.php
+++ b/Command/PopulateCommand.php
@@ -11,6 +11,7 @@ use FOS\ElasticaBundle\IndexManager;
 use FOS\ElasticaBundle\Provider\ProviderRegistry;
 use FOS\ElasticaBundle\Resetter;
 use FOS\ElasticaBundle\Provider\ProviderInterface;
+use Symfony\Component\Console\Helper\ProgressBar;
 
 /**
  * Populate the search index
@@ -109,20 +110,34 @@ class PopulateCommand extends ContainerAwareCommand
      */
     private function populateIndex(OutputInterface $output, $index, $reset, $options)
     {
+
+        /** @var $providers ProviderInterface[] */
+        $providers = $this->providerRegistry->getIndexProviders($index);
+
         if ($reset) {
             $output->writeln(sprintf('<info>Resetting</info> <comment>%s</comment>', $index));
             $this->resetter->resetIndex($index, true);
         }
 
-        /** @var $providers ProviderInterface[] */
-        $providers = $this->providerRegistry->getIndexProviders($index);
-
         foreach ($providers as $type => $provider) {
-            $loggerClosure = function($message) use ($output, $index, $type) {
-                $output->writeln(sprintf('<info>Populating</info> %s/%s, %s', $index, $type, $message));
-            };
+            if (class_exists('Symfony\Component\Console\Helper\ProgressBar')) {
+                $output->writeln(sprintf('<info>Populating</info> %s/%s', $index, $type));
+                $progressBar = new ProgressBar($output, $provider->getTotalObjects());
+                $progressBar->setFormat('debug');
+                $progressBar->start();
+                $loggerClosure = function($number) use ($progressBar) {
+                    $progressBar->advance($number);
+                };
+            } else {
+                $loggerClosure = function($message) use ($output, $index, $type) {
+                    $output->writeln(sprintf('<info>Populating</info> %s/%s, %s', $index, $type, $message));
+                };
+            }
 
+            $options['progress-bar'] = true;
             $provider->populate($loggerClosure, $options);
+
+            if (isset($progressBar)) $progressBar->finish();
         }
 
         $output->writeln(sprintf('<info>Refreshing</info> <comment>%s</comment>', $index));
--- a/Doctrine/AbstractProvider.php
+++ b/Doctrine/AbstractProvider.php
@@ -7,6 +7,7 @@ use Elastica\Exception\Bulk\ResponseException as BulkResponseException;
 use FOS\ElasticaBundle\Persister\ObjectPersisterInterface;
 use FOS\ElasticaBundle\Provider\AbstractProvider as BaseAbstractProvider;
 use FOS\ElasticaBundle\Provider\IndexableInterface;
+use Symfony\Component\Console\Helper\ProgressBar;
 
 abstract class AbstractProvider extends BaseAbstractProvider
 {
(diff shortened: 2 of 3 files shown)


Doctrine/AbstractProvider.php
```diff
@@ -53,6 +54,7 @@ public function populate(\Closure $loggerClosure = null, array $options = array(
         $sleep = isset($options['sleep']) ? intval($options['sleep']) : 0;
         $batchSize = isset($options['batch-size']) ? intval($options['batch-size']) : $this->options['batch_size'];
         $ignoreErrors = isset($options['ignore-errors']) ? $options['ignore-errors'] : $this->options['ignore_errors'];
+        $progresBar = isset($options['progress-bar']) ? boolval($options['progress-bar']) : false;
```
daFish: Should be `$progressBar`.

Doctrine/AbstractProvider.php
```diff
@@ -94,12 +96,14 @@ public function populate(\Closure $loggerClosure = null, array $options = array(
 
             usleep($sleep);
 
-            if ($loggerClosure) {
+            if ($loggerClosure && !$progresBar) {
```
daFish: Same here.

Doctrine/AbstractProvider.php
```diff
@@ -94,12 +96,14 @@ public function populate(\Closure $loggerClosure = null, array $options = array(
...
                 $stepCount = $stepNbObjects + $offset;
                 $percentComplete = 100 * $stepCount / $nbObjects;
                 $timeDifference = microtime(true) - $stepStartTime;
                 $objectsPerSecond = $timeDifference ? ($stepNbObjects / $timeDifference) : $stepNbObjects;
                 $loggerClosure(sprintf('%0.1f%% (%d/%d), %d objects/s %s', $percentComplete, $stepCount, $nbObjects, $objectsPerSecond, $this->getMemoryUsage()));
+            } else if ($loggerClosure && $progresBar) {
```
daFish: Same here.

pmartelletti:
@daFish done.

I've also added support for Propel, that was not included in the previous commit.

Command/PopulateCommand.php
```diff
@@ -109,20 +110,34 @@ protected function execute(InputInterface $input, OutputInterface $output)
      */
     private function populateIndex(OutputInterface $output, $index, $reset, $options)
     {
+
```
merk: Please remove the extra line here

Command/PopulateCommand.php
```diff
@@ -109,20 +110,34 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        /** @var $providers ProviderInterface[] */
+        $providers = $this->providerRegistry->getIndexProviders($index);
```
merk: Why rearrange this variable when it isnt necessary for the reset code?

Command/PopulateCommand.php
```diff
@@ -109,20 +110,34 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
-        $providers = $this->providerRegistry->getIndexProviders($index);
-
         foreach ($providers as $type => $provider) {
-            $loggerClosure = function($message) use ($output, $index, $type) {
-                $output->writeln(sprintf('<info>Populating</info> %s/%s, %s', $index, $type, $message));
-            };
+            if (class_exists('Symfony\Component\Console\Helper\ProgressBar')) {
+                $output->writeln(sprintf('<info>Populating</info> %s/%s', $index, $type));
+                $progressBar = new ProgressBar($output, $provider->getTotalObjects());
+                $progressBar->setFormat('debug');
```
merk: It should be up to the caller of the command to specify the verbosity to get the information they need. Using -vvv will give you debug output.

Command/PopulateCommand.php
```diff
@@ -109,20 +110,34 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            } else {
+                $loggerClosure = function($message) use ($output, $index, $type) {
+                    $output->writeln(sprintf('<info>Populating</info> %s/%s, %s', $index, $type, $message));
+                };
+            }
 
+            $options['progress-bar'] = true;
             $provider->populate($loggerClosure, $options);
+
+            if (isset($progressBar)) $progressBar->finish();
```
merk: PSR-2 standard here please

```php
if (isset($progressBar)) {
    $progressBar->finish();
}
```

Command/PopulateCommand.php
```diff
@@ -109,20 +110,34 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         }
 
-        /** @var $providers ProviderInterface[] */
-        $providers = $this->providerRegistry->getIndexProviders($index);
-
         foreach ($providers as $type => $provider) {
-            $loggerClosure = function($message) use ($output, $index, $type) {
-                $output->writeln(sprintf('<info>Populating</info> %s/%s, %s', $index, $type, $message));
-            };
+            if (class_exists('Symfony\Component\Console\Helper\ProgressBar')) {
```
merk: Please move the creation of the loggerClosure to its own method.

I also think that instead of having the progress-bar option in the Provider we can just pass the necessary information as parameters to the loggerClosure and let it handle everything.

IE, we'd call `$loggerClosure($stepCount, $nbObjects, $stepStartTime)` and then format that information how we want it formatted depending on if ProgressBar is available.

merk: This means we dont need special logic in the Provider at all, if a loggerClosure is present we call it with the same data.

merk: We'll need a $message variable too..

pmartelletti: @merk sound logic, yes. I will try to work on this during the weekend. :smile:

Command/PopulateCommand.php
```diff
@@ -109,20 +110,34 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                $output->writeln(sprintf('<info>Populating</info> %s/%s', $index, $type));
+                $progressBar = new ProgressBar($output, $provider->getTotalObjects());
+                $progressBar->setFormat('debug');
+                $progressBar->start();
+                $loggerClosure = function($number) use ($progressBar) {
+                    $progressBar->advance($number);
+                };
+            } else {
+                $loggerClosure = function($message) use ($output, $index, $type) {
```
stof: use the ProgressHelper when the ProgressBar is not there (it will allow to share most logic, and to have a consistent output)

merk: I've just spent some time implementing this and ProgressHelper is very limited in what it can handle. I dont think its worth it.

merk:
Updated code and sent a new PR #780

Command/PopulateCommand.php
stof: Well, The ProgressHelper is similar to the PRogressBar here. The difference is small, and it has the API we need AFAICT

merk: It isnt capable of handling additional information like what index or types are being populated and it isnt possible to extend to add the additional message parameters because everything is private.

Additionally, to hide the bar, write a message and then redisplay it would mean I have to pass the bar around everywhere in the PopulateCommand.

PR #772, opened 2015-01-19
Typo in manual-provider.md

CedCannes:
Typo in class path

### DIFF
--- a/Resources/doc/cookbook/manual-provider.md
+++ b/Resources/doc/cookbook/manual-provider.md
@@ -8,7 +8,7 @@ index and type for which the service will provide.
 # app/config/config.yml
 services:
     acme.search_provider.user:
-        class: Acme\UserBundle\Search\UserProvider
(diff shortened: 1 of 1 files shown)


merk:
I dont believe this to be a typo. I recommend in this circumstance (of a application's UserBundle) to group classes relating to purpose, not their type. In this case I'd expect to put all FOSElasticaBundle custom classes under the Search\ namespace.

CedCannes:
I agree with you.
I Thought that a typo mistake beacuse in service configuration we have : 
class: Acme\UserBundle\Search\UserProvider

and the namespace of the class UserProvider in example is : 
namespace Acme\UserBundle\Provider;

Search or Provider  ? :)

merk:
You're right - thanks I completely missed the discrepancy :)

merk:
Merged with `c4a2858` into 3.0.x

PR #776, opened 2015-01-21
Fix the handling of the parent mapping in the configuration

### LINKED ISSUE #774: [Regression] The configuration refactoring broke the parent child mapping

stof:
This fixes a regression introduced in 3.0 in #629. The issue is because of the refactoring of https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/629/files#diff-e0950e36ea63dd2c4b5151242670298cL244 where adding the parent in the replacement of typeFields was missing.

Closes #774


stof:
Wrong target. It should go in 3.0.x. I will reopen the PR.

stof:
See #777

PR #780, opened 2015-01-22
Continued work on ProgressBar

merk:
Replaces #770

I did not implement ProgressHelper because it has too many problems to be easily implemented, if someone else wants to have a go they're welcome to attempt it.

### DIFF
--- a/Command/PopulateCommand.php
+++ b/Command/PopulateCommand.php
@@ -101,6 +101,81 @@ class PopulateCommand extends ContainerAwareCommand
     }
 
     /**
+     * @param ProviderInterface $provider
+     * @param OutputInterface $output
+     * @param string $input
+     * @param string $type
+     * @param array $options
+     */
+    private function doPopulateType(ProviderInterface $provider, OutputInterface $output, $input, $type, $options)
+    {
+        $loggerClosure = $this->getLoggerClosure($output, $input, $type);
+
+        $provider->populate($loggerClosure, $options);
+    }
+
+    /**
+     * Builds a loggerClosure to be called from inside the Provider to update the command
+     * line.
+     *
+     * @param OutputInterface $output
+     * @param string $index
+     * @param string $type
+     * @return callable
+     */
+    private function getLoggerClosure(OutputInterface $output, $index, $type)
+    {
+        if (!class_exists('Symfony\Component\Console\Helper\ProgressBar')) {
+            $lastStep = null;
+            $current = 0;
+
+            return function ($increment, $totalObjects) use ($output, $index, $type, &$lastStep, &$current) {
+                if ($increment > $totalObjects) {
+                    $increment = $totalObjects;
+                }
+
+                $currentTime = microtime(true);
+                $timeDifference = $currentTime - $lastStep;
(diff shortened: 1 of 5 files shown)


merk:
An example of what it now looks like:

screen shot 2015-01-22 at 11 14 48 am

stof:
This needs to be rebased

Command/PopulateCommand.php
```diff
@@ -100,6 +101,81 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+                    round(memory_get_usage() / (1024 * 1024)),
+                    round(memory_get_peak_usage() / (1024 * 1024))
+                ));
+            };
+        }
+
+        ProgressBar::setFormatDefinition('normal', " %current%/%max% [%bar%] %percent:3s%%\n%message%");
+        ProgressBar::setFormatDefinition('verbose', " %current%/%max% [%bar%] %percent:3s%% %elapsed:6s%\n%message%");
+        ProgressBar::setFormatDefinition('very_verbose', " %current%/%max% [%bar%] %percent:3s%% %elapsed:6s%/%estimated:-6s%\n%message%");
+        ProgressBar::setFormatDefinition('debug', " %current%/%max% [%bar%] %percent:3s%% %elapsed:6s%/%estimated:-6s% %memory:6s%\n%message%");
```
stof: Should we really change these ?
Btw, I'm not use using a multiline message works well cross-platform for the progress bar

Command/PopulateCommand.php
```diff
@@ -100,6 +101,81 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        $progress = null;
+
+        return function ($increment, $totalObjects) use (&$progress, $output, $index, $type) {
+            if (null === $progress) {
+                $progress = new ProgressBar($output, $totalObjects);
```
stof: here is how you could use the ProgressHelper for that:

```php
$progress = new ProgressHelper();
$progress->start($output, $totalObjects);
```

method names for the check whether we should call finish needs to be adapted too

Command/PopulateCommand.php
merk: I dont see a better way to provide feedback about what we're doing while the progress bar is running. It'd be nice if the default formats provided a message placeholder, even if it was on the same line.

If the ProgressBar shouldnt use multi line messages because of platform support we should submit a change to the docs.

Command/PopulateCommand.php
merk: Theres no way to know where the ProgressHelper's current step is at to know if we should call ->finish() or not. The primary problem though is there is no way to output additional information while the ProgressHelper is displayed without passing it everywhere.

Command/PopulateCommand.php
merk: I changed these so we could hook into the default format detection. I guess we could just set the format to debug and always output everything?

Command/PopulateCommand.php
stof: hmm, indeed, ProgressHelper does not have the getter

PR #782, opened 2015-01-26
Update PopulateCommand.php

allanbrault:
in the case i have 110 objects, it was doing :
Populating abc/Index 90.9% (100/110)
Populating abc/index 181.8% (200/110)

now :
Populating abc/Index 90.9% (100/110)
Populating abc/index 100.0% (110/110)


merk:
Thanks

PR #789, opened 2015-02-02
fix error for empty type configuration

ewgRa:
I use elastica-bundle without configuration types, like this:

```fos_elastica:
    indexes:
        foo:
            types:
                bar:
                bar2:
```

This pull request fix warning:
"PHP Warning:  array_key_exists() expects parameter 2 to be array, null given in .../vendor/friendsofsymfony/elastica-bundle/DependencyInjection/Configuration.php on line 197" with such configuration


merk:
Merged into 3.0.x as cf9f7c6

merk:
Thanks!

PR #793, opened 2015-02-06
Fix "progress bar past the max value" on populate

BourotBenjamin:
Because the progress bar crash if it is set to more than 100%


merk:
@Stof Do you have an opinion on only supporting 2.3 and 2.6+? (eg 2.3.*||~2.6) I dont really want to have to externally track the progress bar's progress just so we dont go over the maximum.

BourotBenjamin:
Un fact, the problem still exist in 2.5.9 ;-)

BourotBenjamin:
So, what do you think ? (As you already merged your commits that only fix 2.4/2.5, why not merging this one ?)

merk:
This is dealt with in #815

Issue #797, opened 2015-02-19
fos:elastica:populate --no-reset fails when use_alias: true


aheinz-sg:
Using ElasticaBundle v3.0.8, with configuration along the lines of:
```YAML
fos_elastica:
    indexes:
        search:
            use_alias: true
            index_name: my_index
            types:
                product:
                    mappings:
                        name: { analyzer: standard }
```
Executing the following command throws an exception
```sh
$ app/console fos:elastica:populate --no-reset
Populating search/product, 32.9% (100/304), 113 objects/s (RAM : current=15Mo peak=17Mo)
Populating search/product, 65.8% (200/304), 368 objects/s (RAM : current=15Mo peak=18Mo)
Populating search/product, 98.7% (300/304), 421 objects/s (RAM : current=18Mo peak=20Mo)
Populating search/product, 100.0% (304/304), 394 objects/s (RAM : current=17Mo peak=20Mo)
Refreshing search
  [Elastica\Exception\ResponseException]        
  IndexMissingException[[my_index] missing]
```

I've chased it down a little and I believe the issue is that [PopulateCommand.php:129](https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/3.0.x/Command/PopulateCommand.php#L129) is not being skipped when --no-reset is used, unlike [PopulateCommand.php:114](https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/3.0.x/Command/PopulateCommand.php#L114).

merk:
the use_alias is generally the problem here, --no-reset is not necessary when it is set but there is a problem with the IndexAlias handling if `my_index` isnt already an alias.

I suggest you set up my_index as an alias to another blank index prior to population. I had forgotten to work out how to handle this and will leave this issue open as a reminder to make the alias handling more robust.

PR #798, opened 2015-02-23
Don't run resetter postPopulate if --no-reset. #797

aheinz-sg:
This prevents my alias from getting deleted. It generally seems reasonable not to call any methods on the resetter with the --no-reset option set.


merk:
Closing because this defect is fixed in 3.1-dev due to changes in the command. If someone encounters an issue with --no-reset and aliases in 3.0 and wants to backport the fix I'll merge it in.

Repeating here: not running postPopulate is not the correct solution to this issue.

PR #799, opened 2015-02-23
Parallelising population commands

aheinz-sg:
Building on PR #797, this lets me execute fos:elastica:populate in parallel
```bash
DOCS=`mongo marketplace --eval 'db.products.count()' --quiet`
PROCS=10
STEP=10
eval echo {0..$DOCS..$STEP} |\
xargs -n1 -P$PROCS app/console fos:elastica:populate --no-reset --batch-size $STEP --limit $STEP --index search --type product --offset
```


merk:
Your proposed changes in #797 and here regarding not running postPopulate unless --no-reset is present doesnt make sense to me: why would you populate a brand new index (which happens when you're using use_alias) and not switch the alias over afterwards?

With your second commit it makes more sense that you want to run populate in parallel, but we're going to have to achieve it another way or reserve this change for 4.0 of the bundle at some point in the future.

aheinz-sg:
At least from my own experimentation with --no-reset, it does not create a new index as you describe -- it continues to push products into my existing alias. Then the resetter postPopulate deletes that index, leaving me with nothing -- hence my first commit.

merk:
It looks the reset problem was resolved in 3.1-dev where we dont run postPopulate for an individual type population: https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Command/PopulateCommand.php#L223

This is because of a limitation that we cannot use a new alias if you specify a --type to the command and it has to repopulate in place. Post population should not run in this instance but that is independent of supplying --no-reset. If you adjust #797 to stop post population when you're *only* populating a type I'll be happy to merge that into 3.0.x as a bugfix.

As for the parallelising of population I think I want to go with a different approach where we break out operations into independent commands that will allow better scripting/customisation. I envisage:

* create_alias: creates a new alias to operate on that isnt live
* populate_type: runs population on an individual type. This command type be written in a way to support different strategies for splitting workloads. offset/limit are not enough because many users use a `where primary_key > last_key` approach instead of offset/limit.
* populate: just runs populate_type for each type in an index/all indexes
* switch_alias: takes the alias returned by create_alias and switches it live

populate_type and populate would take a custom --alias parameter to tell the commands to work on a separate alias instead of on the live index..

aheinz-sg:
I like the direction of the 3.1.x reset changes, so I've closed #797.

I'm all for create_alias; being able to use 'index_read' and 'index_write' aliases would really give us a lot of flexibility in terms of how we attack reindexing. We do a lot of live updating of our index currently and the second populate starts, our live updating is sending data to an index that's going to be imminently deleted!

Although I don't disagree with the direction of populate_type, it seems like a modest amount of work.
Aside from your proposed addition of strategies, I don't see a lot of difference between populate_type and populate --index x --type y, which is what I used in my example above. I went with offset/limit since the --offset option already existed. At the very least, this patch allows me to easily scale horizontally across all the processors on my machine, or additional machines in the cluster. Maybe give folks one option to use before they get an entire menu? (Minus the first commit, of course.)

merk:
We encounter the same problem and I havent come up with a solution except thinking about storing what we miss while reindexing and go back and reindex those objects at the end of the process as well.

My problem is the :populate command does too many things and its set of options only work in certain cases. I will spend some time this weekend playing around with splitting it up and send a PR for you to take a look at.

aheinz-sg:
Recreating PR against 3.1.0

PR #739, opened 2014-11-14
adding support for phpcr

mosch:
I wanted support for the phpcr within the FOSElasticaBundle so I wrote this prototype of phpcr support. 

I think it's not complete yet but i would love to get some feedback on this, to know whats missing or what may fail. At the moment I think at least a custom provider class is missing, I'm working on that now.

@dbu would love to see your feedback here.

Also FriendsOfSymfony/FOSElasticaBundle#729 may be in conflict with this and should get merged in frist place.

PR #805, opened 2015-03-02
Adding PHPCR support

webda2l:
Follow https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/739

### DIFF
--- a/Doctrine/PHPCR/ElasticaToDocumentTransformer.php
+++ b/Doctrine/PHPCR/ElasticaToDocumentTransformer.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace FOS\ElasticaBundle\Doctrine\PHPCR;
+
+use Doctrine\ODM\PHPCR\DocumentManager;
+use Doctrine\ODM\PHPCR\Query\Builder\QueryBuilder;
+use Elastica\Result;
+use FOS\ElasticaBundle\Doctrine\AbstractElasticaToModelTransformer;
+use Doctrine\ORM\Query;
+
+use FOS\ElasticaBundle\Transformer\ElasticaToModelTransformerInterface;
+
+/**
+ * Maps Elastica documents with Doctrine objects
+ * This mapper assumes an exact match between
+ * elastica documents ids and doctrine object ids
+ */
+class ElasticaToDocumentTransformer extends AbstractElasticaToModelTransformer
+{
+    /**
+     * Fetch objects for theses identifier values
+     *
+     * @param array $identifierValues ids values
+     * @param Boolean $hydrate whether or not to hydrate the objects, false returns arrays
+     * @return array of objects or arrays
+     */
+    protected function findByIdentifiers(array $identifierValues, $hydrate)
+    {
+        return iterator_to_array($this->registry
+            ->getManagerForClass($this->objectClass)
+            ->getRepository($this->objectClass)
+            ->findMany($identifierValues));
+    }
+}
--- a/Doctrine/PHPCR/ElasticaToModelTransformer.php
+++ b/Doctrine/PHPCR/ElasticaToModelTransformer.php
@@ -2,20 +2,14 @@
 
 namespace FOS\ElasticaBundle\Doctrine\PHPCR;
 
-use Doctrine\ODM\PHPCR\DocumentManager;
-use Doctrine\ODM\PHPCR\Query\Builder\QueryBuilder;
-use Elastica\Result;
 use FOS\ElasticaBundle\Doctrine\AbstractElasticaToModelTransformer;
-use Doctrine\ORM\Query;
-
-use FOS\ElasticaBundle\Transformer\ElasticaToModelTransformerInterface;
 
 /**
  * Maps Elastica documents with Doctrine objects
  * This mapper assumes an exact match between
  * elastica documents ids and doctrine object ids
  */
-class ElasticaToDocumentTransformer extends AbstractElasticaToModelTransformer
+class ElasticaToModelTransformer extends AbstractElasticaToModelTransformer
 {
     /**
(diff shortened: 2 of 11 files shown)


lsmith77:
looks sensible from a quick inspection

/cc @dbu

Doctrine/PHPCR/ElasticaToDocumentTransformer.php
```diff
@@ -0,0 +1,41 @@
...
+                    ->getManager()
+                    ->findMany(null, $identifierValues)
+            );
+        }
+
+        // General case
+        return iterator_to_array(
+            $this->registry
+                ->getManager()
+                ->getRepository($this->objectClass)
```
dbu: in terms of phpcr-odm, this means that objects of this class or extending classes will be found. which sounds like a good idea to me. (so just to make sure you are aware of that)

Doctrine/PHPCR/ElasticaToDocumentTransformer.php
```diff
@@ -0,0 +1,41 @@
...
+     *
+     * @param array $identifierValues ids values
+     * @param Boolean $hydrate whether or not to hydrate the objects, false returns arrays
+     * @return array of objects or arrays
+     */
+    protected function findByIdentifiers(array $identifierValues, $hydrate)
+    {
+        // Special case where model is interface. But maybe remove because negative performance impact
+        $reflectionClass = new \ReflectionClass($this->objectClass);
+        if ($reflectionClass->isInterface()) {
```
lsmith77: maybe http://php.net/manual/en/function.interface-exists.php is more efficient .. or get_declared_interfaces() .. though the later I doubt is more efficient.

merk:
Not knowing much about PHPCR, I'll defer to @dbu and @lsmith77's evaluations.

I am nervous about merging as is: without any integration tests it isnt possible for me to know if any changes will break this code. I realise there are few tests for other backends but we must start somewhere.

dbu:
it looks fine, but a few tests would sure make sense. @merk can you point @webda2l to some examples how to do this?

a short note in the doc would also make sense so people actually detect this exists.

Doctrine/PHPCR/ElasticaToDocumentTransformer.php
webda2l: Yes I'm aware, thks

Doctrine/PHPCR/ElasticaToDocumentTransformer.php
webda2l: Maybe I'm wrong, but theses methods doesn't seems fit the needs... oO

If I want avoid this Reflection use, but allow such ElasticableInterface use for multiple models in foselastica settings, I could do differently however.
Check in a higher level, during foselastica init configuration, if the model value is an interface case and pass this information to the ElasticaToModelTransformer by example.
But see if its has plus-value.
Or I can also remove this special case and do it only for my project.

Doctrine/PHPCR/ElasticaToModelTransformer.php
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @param array $identifierValues ids values
+     * @param Boolean $hydrate whether or not to hydrate the objects, false returns arrays
+     * @return array of objects or arrays
+     */
+    protected function findByIdentifiers(array $identifierValues, $hydrate)
+    {
+        // Special case where model is interface
+        $reflectionClass = new \ReflectionClass($this->objectClass);
+        if ($reflectionClass->isInterface()) {
```
dbu: i think [interface_exists](http://php.net/manual/en/function.interface-exists.php) is exactly what you want here.

merk:
As a starting point: duplicating some of the code under Tests/Functional that relates to the ORM kernel configuration would be great.

webda2l:
This is limited in written tests.. but this PR could be RFR.

dbu:
this looks good to me. what about merging it and then trying to bootstrap some more thorough functional testing for the whole bundle?

Buri:
Any progress here?

dbu:
oh, there was no comment after the updated commits so i got no notification. @merk this looks good to me (but i did not run the code, just glance over it). imho we can merge this.

dymissy:
Any news about the PR?

dbu:
@merk are you still active on this bundle?

dymissy:
If could help someone, I tried the PR code on my environment and it works fine. Contents are indexed correctly. I just have to find a solution for multi-language documents but if the PR will be approved I could try to add a new one in the next month.

merk:
I am not particularly active at the moment because my current projects dont have searching capabilities and I cant spare the time.

I'm happy for other FOS team members to help out if they have time. I'd love to do some major refactoring since I'm not happy with the way the bundle is used but it'll have to wait.

dbu:
okay, i see. its the same for me. i was only on this branch because of PHPCR. i guess we can merge this PR, ok?

merk:
My only concern is taking on more code becomes a liability if noone will maintain it (i dont use PHPCR). Other than that, happy to merge

dbu:
i can help with questions/problems about the PHPCR-ODM integration.

dbu:
@webda2l thanks a lot!
i notice we miss documentation. @dymissy or @webda2l could one of you do a pull request to add this to https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Resources/doc/types.md#realtime-selective-index-update ? (and maybe check the doc for other mentions of doctrine orm and add phpcr odm as another option there?)

dymissy:
@dbu I just created a new PR here https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/966

While I'm using the library I will try to update the doc with some example using PHPCR.

PR #806, opened 2015-03-03
Add a reindex command to update index mapping without downtime

progrupa:
This pull request creates a ReindexCommand which allows to update index mapping without downtime.
It creates a new index, updates it's mappings, populates it with the content of the old index. Then it switches the alias used and updates changes made during the reindex process.
Only works with aliased indices.

### DIFF
--- a/Command/LoggerClosureHelper.php
+++ b/Command/LoggerClosureHelper.php
@@ -1,0 +1,80 @@
+<?php
+/**
+ * Created by PhpStorm.
+ * User: dominikkasprzak
+ * Date: 27/02/15
+ * Time: 11:16
+ */
+
+namespace FOS\ElasticaBundle\Command;
+
+
+use Symfony\Component\Console\Helper\ProgressBar;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class LoggerClosureHelper
+{
+    /**
+     * Builds a loggerClosure to be called from inside the Provider to update the command
+     * line.
+     *
+     * @param OutputInterface $output
+     * @param string $index
+     * @param string $type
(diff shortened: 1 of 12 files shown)


merk:
Can you please talk a bit more about why you copy the index to a new one? Are you using external tools to populate index data on top of Elastica/FOSElasticaBundle that you need to keep across new populations?

merk:
Overall I think I like where this is headed. I'd like to see the copying of documents moved out into its own general service that only handles copying, and if you can fix up the CS of the pull (missing copyright headers at the top, PSR2 issues with some of the files, use PHPUnit standard method names).

I also noticed that I split out the logger closure stuff into its own service a few minutes ago but missed that you'd done that - sorry.

progrupa:
Well, my use case is that I use ES as a standalone storage, therefore I need to be able to change index mapping information without losing data. I have no source to populate it from.

I'll work on the issues You've mentioned.

merk:
What are you reindexing from then? Wouldn't it make more sense to have independent indexes for different data sources and search combined indexes?

progrupa:
It depends on your use case.
For me, ElasticSearch is a primary data source, and this command allows me to modify index/type mappings without losing any data.

merk:
I suggest you contribute functionality included here for copying data in indexes to the Elastica project because this requirement is not specific to FOSElasticaBundle and its users and may be useful for other people using Elastica that do not use symfony.

You can then leverage that capability in your own project to modify index/types without losing data. I'm not sure FOSElasticaBundle should have much responsibility here (you can use the services defined by the bundle once Elastica has the feature)

PR #808, opened 2015-03-05
Enable use of hints when transforming results to entities

progrupa:
Useful when using extensions like translatable


merk:
Thanks! This looks like a good start but I dont like the idea of adding hints as a concept for all elastica_to_model transformers because other backends do not support such a feature.

I would like to see an exception thrown if hints are configured for a backend that doesnt support them.

progrupa:
Sure thing, there You go.

merk:
Sorry, I should have been more specific - I'd prefer to see the validation occur in the Configuration class, not in the Extension. It can be done using this approach: http://symfony.com/doc/current/components/config/definition.html#validation-rules

progrupa:
No problem, moved it to the Configuration class.

merk:
Could you please add a cookbook entry that explains the use of hints with an example?

Thanks. Once thats done this should be ready to merge into 3.2-dev.

progrupa:
Should it be included in an existing section, or would You rather have it in a separate cookbook entry?

merk:
I think a separate cookbook entry makes more sense here.

progrupa:
Added the cookbook entry.
Also, renamed the 'class' parameter to 'value', since query hints accept arbitrary values, not only class names like the Query::HINT_CUSTOM_TREE_WALKERS.

XWB:
Thank you.

PR #816, opened 2015-03-11
SliceFetcher

merk:
Takes #725 and fixes merge conflicts and tidys code a tiny bit


merk:
@Gnucki Sorry it took so long - if you have a chance to take a look at the AbstractProvider (I moved the sliceFetcher logic into a new method instead) to make sure I havent broken anything that you're relying on that'd be awesome.

Gnucki:
I will try to test it tomorrow!

merk:
@Gnucki I just realised that the MongoDB slicefetcher has a different behaviour than the one it replaces. I dont use MongoDB so im not sure if its a big deal or not?

I think it would be better to default to using the same limit/skip query that the Provider uses with an option to enable a more efficient method until we deprecate the old way?

Gnucki:
The old way has a really big problem because the more documents you have in your collection the more the indexation slows for each slice. It seems that around 1 000 000 documents the time to retrieve and index a slice is so slow that the process is failling (timeout).

For me, it's an enough serious problem to enable the new slice fetching by default. Moreover, the new slice fetching for mongodb retrieves exactly the same slices as the old one but in a constant time for each slice. I use it in prod environment.

However, you are the one who best understand what is good for all users of this bundle so do what you think is better!

And sorry, I had so much work this week that I didn't have time to test it. I will try to do it on the early days of next week.

Gnucki:
Tested on master branch. This is working.

PR #820, opened 2015-03-12
General code qa

merk:
I've just spent a bit of time going over scrutinizer and fixing issues that were easy pickings.


Doctrine/Listener.php
```diff
@@ -84,7 +84,7 @@ public function __construct(
         $this->objectPersister = $objectPersister;
         $this->propertyAccessor = PropertyAccess::createPropertyAccessor();
 
-        if ($logger) {
+        if ($logger && $this->objectPersister instanceof ObjectPersister) {
```
stloyd: You say that's interface, yet you check for class implementing it...

merk:
merged into 3.1.x

PR #822, opened 2015-03-12
Doctrine listener exception fix

vkartaviy:
After update I became receiving following exception:
```
[Symfony\Component\Debug\Exception\ContextErrorException]
  Catchable Fatal Error: Argument 4 passed to FOS\ElasticaBundle\Doctrine\Listener::__construct() must implement interface
   Psr\Log\LoggerInterface, string given, called in /var/www/Kolektado/app/cache/dev/appDevDebugProjectContainer.php on li
  ne 1727 and defined in /var/www/Kolektado/vendor/friendsofsymfony/elastica-bundle/Doctrine/Listener.php line 79
```


merk:
Thanks - this is already fixed

PR #836, opened 2015-03-27
Forced usage of symfony/options-resolver ~2.6

xavismeh:
This requirement is mandatory because of https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/3.1.x/Provider/AbstractProvider.php#L111

:panda_face:


merk:
Thanks, but this isnt the correct solution: this bundle will maintain compatibility with 2.3 until it is EOL.

I've pushed a new PR that should hopefully resolve this issue.

PR #838, opened 2015-03-30
Fix an error when batch size is empty

huitiemesens:
if batch size is null or empty , doPopulate will loop forever on the
first type indexing

### DIFF
--- a/Command/PopulateCommand.php
+++ b/Command/PopulateCommand.php
@@ -90,8 +90,10 @@ class PopulateCommand extends ContainerAwareCommand
         $index = $input->getOption('index');
         $type = $input->getOption('type');
         $reset = !$input->getOption('no-reset');
+        $batchSize = $input->getOption('batch-size');
+        
         $options = array(
-            'batch_size' => $input->getOption('batch-size'),
+            'batch_size' => !empty( $batchSize ) ? $batchSize : 1,
             'ignore_errors' => $input->getOption('ignore-errors'),
             'offset' => $input->getOption('offset'),
             'sleep' => $input->getOption('sleep')


Command/PopulateCommand.php
```diff
@@ -91,7 +91,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $type = $input->getOption('type');
         $reset = !$input->getOption('no-reset');
         $options = array(
-            'batch_size' => $input->getOption('batch-size'),
+            'batch_size' => ( !empty($input->getOption('batch-size') ) ) ? $input->getOption('batch-size') : 1,
```
stof: this is not compatible with PHP 5.3 and 5.4 where ``empty`` can only be used on variables

huitiemesens: i updated for php 5.3 / 5.4 compatibility

Command/PopulateCommand.php
```diff
@@ -90,8 +90,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $index = $input->getOption('index');
         $type = $input->getOption('type');
         $reset = !$input->getOption('no-reset');
+        $batch_size = $input->getOption('batch-size');
```
stof: camelCase name please

Command/PopulateCommand.php
```diff
@@ -90,8 +90,10 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+        
         $options = array(
-            'batch_size' => $input->getOption('batch-size'),
+            'batch_size' => ( !empty( $batch_size ) ) ? $batch_size : 1,
```
stof: parenthesis around the condition are not necessary

huitiemesens:
okay i cleaned and squashed intermediate commits

webda2l:
It can be a fix but not the right in my opinion.

I think @merk want default value for batch_size from https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Provider/AbstractProvider.php#L108. There is something to investigate with its refactoring https://github.com/FriendsOfSymfony/FOSElasticaBundle/commit/3bb2f384babf9f72461ddab6f332c7989376ad81

huitiemesens:
@webda2l you re probably right , all i can say right now is fos:elastica:populate can't be run fully without setting manually a default value.

merk:
Thanks for the PR - this should be fixed in the 3.1 branch with b6e01cd332ffa29ab354c9d882b5686f698c1d06

PR #846, opened 2015-04-02
Fixes wrong count for queries with joins in Doctrine ORM Provider

### LINKED ISSUE #844: Object count is wrong when populating with Doctrine ORM provider and custom query_builder_method
The method `\FOS\ElasticaBundle\Doctrine\ORM\Provider::countObjects` doesn't deal with joins. Joins shall be removed from QueryBuilder when counting.

When custom query builder uses joins the method reports all rows returned by the raw query which is usually a lot more. This causes the population to take more time because the object count is used for looping/batching in `\FOS\ElasticaBundle\Doctrine\AbstractProvider::doPopulate`.

pinkeen:
Partially fixes #844 by adding distinct to the count query. With this fix, custom queries can use joins (though not fetch-joins) without breaking populate.

### DIFF
--- a/Doctrine/ORM/Provider.php
+++ b/Doctrine/ORM/Provider.php
@@ -61,7 +61,7 @@ class Provider extends AbstractProvider
         $rootAliases = $queryBuilder->getRootAliases();
 
         return $qb
-            ->select($qb->expr()->count($rootAliases[0]))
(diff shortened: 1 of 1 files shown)


pinkeen:
Okay, upon consideration I think this might break population completely when using fetch-joined query. :)

XWB:
Alright, closing this PR.

PR #852, opened 2015-04-07
Fix logger closure

ebuildy:


davidfuhr:
This does *not* work for me elastica-bundle ~3.1. With which version did you test this?

The comment in `FOS\ElasticaBundle\Provider\ProviderInterface` reads:

```php
     * The closure can expect 2 or 3 arguments:
     *   * The step size
     *   * The total number of objects
     *   * A message to output in error conditions (not normally provided)
```

So if we populate only one user the call should be:

```php
$loggerClosure(1, 1);
```

I extended and refactored the custom provider documentation in #872.

PR #857, opened 2015-04-10
Add fos:elastica:populate --limit option.

aheinz-sg:
I've reimplemented this PR (previous #799) against 3.1.0 to take advantage of the populateIndex() separation (that did not exist in 3.0.0). The default FOS\ElasticaBundle\Doctrine\MongoDB\SliceFetcher uses a `where primary_key > last_key` style of iteration that works great, but the skip($offset) call becomes prohibitive past a few million documents, slowing populate to a crawl (making --offset useless). We happen to have a sequentially indexed integer that we can use instead of the hash primary key, which lets me add the offset to the integer key to avoid the expensive skip call.

``` php
/**
 * {@inheritdoc}
 */
class SliceFetcher implements SliceFetcherInterface
{
    /**
     * {@inheritdoc}
     */
    public function fetch($queryBuilder, $limit, $offset, array $previousSlice, array $identifierFieldNames)
    {
        if (!$queryBuilder instanceof Builder) {
            throw new InvalidArgumentTypeException($queryBuilder, 'Doctrine\ODM\MongoDB\Query\Builder');
        }

        $lastObject = array_pop($previousSlice);
        if (!$lastObject) {
            $lastObject = $queryBuilder
                ->limit(1)
                ->sort(array('humanReadableId' => 'asc'))
                ->getQuery()
                ->getSingleResult();
            $lastObjectId = $lastObject->getHumanReadableId() + $offset;
        } else {
            $lastObjectId = $lastObject->getHumanReadableId();
        }

        return $queryBuilder
            ->field('humanReadableId')
                ->gt($lastObjectId)
            ->limit($limit)
            ->sort(array('humanReadableId' => 'asc'))
            ->getQuery()
            ->execute()
            ->toArray();
    }
}
```

There are gaps in the sequence, so I iterate over the span of integer ids instead of the number of documents in mongo. This amounts to around 3% inefficiency as some documents are submitted more than once. We more than make that back when we toss multiple processors at the problem:

``` bash
#!/bin/bash -ex
exec 1> >(tee /tmp/reindex.log) 2>&1
echo $$ >/tmp/reindex.pid
MONGO='ssh mongo1 mongo marketplace --quiet'
APP="sudo -u www-data php -d memory_limit=1G /usr/share/nginx/www/sites/sgmarketplace/app/console"
#DOCS=`echo 'db.products.count()' | $MONGO`
MIN=`echo 'db.products.find({humanReadableId:{$exists:true}}).sort({humanReadableId:1}).limit(1)[0].humanReadableId.valueOf()' | $MONGO`
MAX=`echo 'db.products.find().sort({humanReadableId:-1}).limit(1)[0].humanReadableId.valueOf()' | $MONGO`
DOCS=`expr $MAX - $MIN`
BATCH=1000
PROCS=`grep -c processor /proc/cpuinfo`
LIMIT=100000
$APP fos:elastica:reset
$APP fos:elastica:populate --index search --type store
eval echo {0..$DOCS..$LIMIT} |\
xargs -n1 -P$PROCS $APP fos:elastica:populate --no-reset --batch-size $BATCH --limit $LIMIT --index search --type product --offset
```

Final results, on a four processor machine, my populate time is cut down to roughly five hours from sixteen. We'll throw even more processors at it in production. ;)

### DIFF
--- a/Command/PopulateCommand.php
+++ b/Command/PopulateCommand.php
@@ -55,6 +55,7 @@ class PopulateCommand extends ContainerAwareCommand
             ->addOption('type', null, InputOption::VALUE_OPTIONAL, 'The type to repopulate')
             ->addOption('no-reset', null, InputOption::VALUE_NONE, 'Do not reset index before populating')
             ->addOption('offset', null, InputOption::VALUE_REQUIRED, 'Start indexing at offset', 0)
+            ->addOption('limit', null, InputOption::VALUE_REQUIRED, 'Populate at most this many documents', 0)
             ->addOption('sleep', null, InputOption::VALUE_REQUIRED, 'Sleep time between persisting iterations (microseconds)', 0)
             ->addOption('batch-size', null, InputOption::VALUE_REQUIRED, 'Index packet size (overrides provider config option)')
             ->addOption('ignore-errors', null, InputOption::VALUE_NONE, 'Do not stop on errors')
@@ -97,6 +98,9 @@ class PopulateCommand extends ContainerAwareCommand
         );
         if ($input->getOption('batch-size')) {
             $options['batch_size'] = (int) $input->getOption('batch-size');
+        }
+        if ($input->getOption('limit')) {
+            $options['limit'] = (int) $input->getOption('limit');
         }
 
         if ($input->isInteractive() && $reset && $input->getOption('offset')) {
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -514,6 +514,7 @@ class Configuration implements ConfigurationInterface
                             ->defaultValue($this->debug)
                             ->treatNullLike(true)
                         ->end()
+                        ->scalarNode('limit')->end()
                         ->scalarNode('query_builder_method')->defaultValue('createQueryBuilder')->end()
                         ->scalarNode('service')->end()
                     ->end()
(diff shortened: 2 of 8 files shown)


aheinz-sg:
I've uncovered a bug when the offset gets larger than the total number of documents. I think we're honestly going to abandon use of fos:elastica:populate entirely in favor of a message bus.

merk:
If you dont mind I'd like to leave this PR open for the time being. I agree with you that fos:elastica:populate is not a very good solution and we need to provide better tools.

Over the next few weeks I hope to contribute some of FOSElastica's code back to Elastica and while thats happening provide better tools for large installs.

borNfreee:
@merk what tools you are talking about? Do you have WIP PRs?
I find a way to populate 50 x 10^6 records to Elastica...

webda2l:
This could be nice to merge this PR as current existing solution.
Personnaly, I use a fork with this PR + PHPCR support

merk:
@webda2l I need to find some time to add some integration tests to the ORM/MongoDB adapters - I'm trying very hard not to merge code without tests because breaking BC is annoying for users.

@borNfreee I'm not sure that using an ORM for populating that many records makes sense. Writing a custom script that keeps operates on the data and creates Elastica documents directly is probably a much speedier approach (or even just builds json strings to POST against ElasticSearch).

aheinz-sg:
Rebased PR, added functional tests and small display improvement. I've fixed my offset bug by updating the code in my slice fetcher to:
```php
            ->field('humanReadableId')
                ->gt($lastObjectId)
                ->lte($lastObjectId + $limit)
```
Which brings the count back in line so I'm not skipping documents.

PR #860, opened 2015-04-17
Count the number of results for the query directly

kayue:
You don't have to execute the entire query to just count the number of result.


merk:
Thanks!

PR #870, opened 2015-04-24
Fixed parameter name to kernel.environment

davidfuhr:
The parameter %kernel.env% does not exist. It throws the error message 'You have requested a non-existent parameter "kernel.env". Did you mean this: "kernel.environment"?'


merk:
Thanks!

PR #871, opened 2015-04-27
Fix propel usage

Seldaek:
Without this fix I got:

```
  [Symfony\Component\OptionsResolver\Exception\InvalidOptionsException]
  The options "clear_object_manager", "debug_logging", "ignore_errors", "offset", "query_builder_method", "sleep" do
  not exist. Known options are: "batch_size", "indexName", "skip_indexable_check", "typeName"
```

### DIFF
--- a/Propel/Provider.php
+++ b/Propel/Provider.php
@@ -43,6 +43,23 @@ class Provider extends AbstractProvider
     /**
      * {@inheritDoc}
      */
+    protected function configureOptions()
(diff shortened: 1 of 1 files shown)


merk:
Hrm, when I made these changes I didnt think about propel, but if the base Provider and anything that calls it will pass these options maybe it would be better to add them as options to the base Provider.

I've done that in #873

PR #872, opened 2015-04-27
Refactor custom provider doc

davidfuhr:
The term "manual provider" was misleading. I did not want to rename the
file to avoid breaking existing links.

Example for index configuration added, improved markup, ordered imports
alphabetically, fixed error with usage of $loggerClosure.

I also added a loop to the provider to illustrate a more "realistic" use
case.

### DIFF
--- a/Resources/doc/cookbook/manual-provider.md
+++ b/Resources/doc/cookbook/manual-provider.md
@@ -1,33 +1,51 @@
-Manual provider
+Custom provider
 ===============
 
-Create a service with the tag "fos_elastica.provider" and attributes for the
-index and type for which the service will provide.
+Usually the index gets populated from the database using the `orm` or `propel`
+provider. But sometimes you might want to index stuff not stored in the
+database. For example you want to index users from another backend or files
+from some storage or some remote resources.
+
+First you need to remove the `provider` stuff from the `config.yml`:
 
 ```yaml
 # app/config/config.yml
+fos_elastica:
+    indexes:
+        app:
+            types:
+                user:
+                    mappings:
+                        id: ~
+                        username: ~
+```
+
+Next create a service with the tag `fos_elastica.provider` and attributes for
+the index and type for which the service will provide.
+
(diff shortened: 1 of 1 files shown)


merk:
Thanks for working on this. It is headed in a good direction however I'd prefer to see the implementation use the ObjectPersister service (which may require changes to the way FOSElasticaBundle is configured for this kind of setup - from memory we dont configure this without a persistence key in the config) and the base AbstractProvider class - they provide functionality that allows developers to use more functionality provided by this bundle while configuring indexes like transformations and handling is_indexable calls.

Resources/doc/cookbook/manual-provider.md
```diff
@@ -35,22 +53,43 @@ class UserProvider implements ProviderInterface
...
-        if ($loggerClosure) {
-            $loggerClosure('Indexing users');
+        // some logic to load our users from somewhere
+
+        $usersCount = count($users);
+        foreach ($users as $user) {
+            if ($loggerClosure) {
+                // first argument is step size, second argument total count
+                // will display a nice progress bar while populating
+                $loggerClosure(1, $usersCount');
```
korotovsky: There is typo `'`

davidfuhr: Thanks! Fixed it.

yellow1912:
I wonder if I can skip the whole persistence definition section? 

In my case I want to index some static content which is not store in any ORM model so the persistence definition is useless for me. I see in this document you do not mention the persistence config but in my case if I remove that I will run into an error with missing index for array.

merk:
It should be possible to proceed without the persistence configuration. Errors without it are a bug.

DjangoFR:
Is custom provider the "right place" to disable/enable some doctrine filters before populating?

$this->em->getFilters()->disable('doctrinefilter');

XWB:
Closing this PR as it it outdated. Feel free to rebase and open a new PR.

PR #875, opened 2015-04-29
[DX] Use consistent index name in documentation

davidfuhr:
I found it confusing that there are different index names used
throughout the documentation. I think using the index name "app"
consistently everywhere makes the docs easier to understand.


merk:
Thanks

PR #877, opened 2015-04-29
[DX] Improve configuration validation

davidfuhr:
The two nodes elastica_to_model_transformer.hydrate and
elastica_to_model_transformer.ignore_missing must be boolean values but
this was not validated in the configuration.

This early validation provides better feedback to the developer who
supplied an invalid configuration.


merk:
Thanks. Cherry picked into 3.0.x as e268698

PR #881, opened 2015-05-05
Update manual-provider.md within right service location

ggnet:
Fixed file name


merk:
The primary symfony-standard application file still goes in app/config/config.yml from what I can tell.

PR #885, opened 2015-05-20
Update Elastica dependency

XWB:


mattjanssen:
We, too, would like to use Elastica 2.0 in order to use Elasticsearch 1.5.

From what I can tell, there are very few breaking changes: https://github.com/ruflin/Elastica/releases/tag/2.0.0

merk:
The condition change is wrong here, we need to disallow 2.1-dev. I've released v3.1.4 and v3.0.10 with Elastica 2.0 allowed.

mattjanssen:
:+1: @merk !

XWB:
:+1:

Issue #887, opened 2015-05-29
RFC: add new event to modify collection of objects when populating database


borNfreee:
Hi,

This bundle has useful `POST_TRANSFORM` event where new field could be added to the `Document`. I need to query database to fetch the value and add it as a new field. So there are 100 additional SQL queries when 100 records are populated.

Of course, it would be greate to make 1 SQL query with `SELECT ... WHERE id IN(1,2,3,4...)` and then update each entity with fetched values.
So i suggest to add an event before each Entity is transformed to elastic Document, here:

``` php
// elastica-bundle/Persister/ObjectPersister.php

public function insertMany(array $objects)
{
    $documents = array();

    // <--------- Here I need a way to modify collection
    $event = new TransformCollectionEvent(objects);
    $this->dispatcher->dispatch(TransformEvent::TRANSFORM_COLLECTION, $event);
    $objects = $event->getObjects();
    // end of new code

    foreach ($objects as $object) {
        $documents[] = $this->transformToElasticaDocument($object);
    }
    try {
        $this->type->addDocuments($documents);
    } catch (BulkException $e) {
        $this->log($e);
    }
}
```

What do you think? If it's ok, I can send a PR.
Please advice.

merk:
The proposed location of the event isnt specific to transformation, i'd create events for each of the *many functions in the Persister - it will allow other operations on objects being added and help moving towards singular service instances rather than an instance per type.

INSERT_OBJECTS, REPLACE_OBJECTS, etc

PR #888, opened 2015-05-29
[WIP] Add PersistingEvent and dispatch INSERT_OBJECTS event when populating

borNfreee:
Implementation for #887
This is a WIP PR. Just want to be on the same page with you @merk 
Please provide your feedback and I'll continue.

TODO:
- [x] create REPLACE_OBJECTS event and dispatch it in appropriate places
- [x] add unit tests for dispatched events
- [ ] update documentation

### DIFF
--- a/DependencyInjection/FOSElasticaExtension.php
+++ b/DependencyInjection/FOSElasticaExtension.php
@@ -453,6 +453,7 @@ class FOSElasticaExtension extends Extension
             'indexName' => $indexName,
             'typeName' => $typeName,
         )));
+        $providerDef->addMethodCall('setEventDispatcher', array(new Reference('event_dispatcher')));
         $container->setDefinition($providerId, $providerDef);
 
         return $providerId;
--- a/DependencyInjection/FOSElasticaExtension.php
+++ b/DependencyInjection/FOSElasticaExtension.php
@@ -417,7 +417,7 @@ class FOSElasticaExtension extends Extension
         foreach ($arguments as $i => $argument) {
             $serviceDef->replaceArgument($i, $argument);
         }
-
+        $serviceDef->addMethodCall('setEventDispatcher', array(new Reference('event_dispatcher')));
         $container->setDefinition($serviceId, $serviceDef);
 
         return $serviceId;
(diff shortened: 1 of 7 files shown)


merk:
I'd prefer to only dispatch in a single place - not having to trigger the same event in multiple places, otherwise I'm okay with the direction.

borNfreee:
> I'd prefer to only dispatch in a single place - not having to trigger the same event in multiple places

What do you mean? There are at least 2 separate places with `->insertMany()` function  where this event should be dispatched (this is for populate command)
* for [Doctrine](https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/a59f2015b49a1b9fc8906b39e724321fe45c9c91/Doctrine/AbstractProvider.php#L95)
* for [Propel](https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/a59f2015b49a1b9fc8906b39e724321fe45c9c91/Propel/Provider.php#L32) 

And the same event should be dispatched when collection of objects [are inserted in the Listener](https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/a59f2015b49a1b9fc8906b39e724321fe45c9c91/Doctrine/Listener.php#L147)

merk:
Thats my point - Dispatching an event before calling insertMany() in 3 places doesnt make sense. The responsibilities of the services that make these calls are not for transformation.

Making the call in the insertMany() call is more appropriate, IMO. https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Persister/ObjectPersister.php#L118

borNfreee:
> Making the call in the insertMany() call is more appropriate

Totally agree, it was [my initial thought](https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/887).
Updated code, added unit tests. Please review.

FYI: tests are failed and seems like this is because of newest symfony 2.7 deprecated messages.

borNfreee:
ping @merk

borNfreee:
ping again @merk

borNfreee:
@merk could you please review this PR or close it if you think this is not something this bundle need?

XWB:
@borNfreee I can merge this PR is if you can update it with master.

XWB:
Closing this PR as nobody really needs it.

PR #902, opened 2015-06-26
adding cookbook for populate events

oleg-andreyev:
Finally I've found some time to write cookbook about `Populate Events`
#744 #814


XWB:
Thanks.

PR #911, opened 2015-07-21
Update settings when resetting a specific type

### LINKED ISSUE #853: Settings not used when populating a specific type
I am using FOSElastica 3.0.9 and I have a setup using aliases where I create an index directly
   curl -XPUT 'http://localhost:9200/index_v1'
then set aliases to point at that index for the different environments
    curl -XPOST 'http://localhost:9200/_aliases' -d '{"actions" : [{ "add" : { "index" : "index_v1", "alias" : "index_prod" } }    ]}'
    curl -XPOST 'http://localhost:9200/_aliases' -d '{"actions" : [{ "add" : { "index" : "index_v1", "alias" : "index_dev" } }    ]}'

I then populate each type individually in order to avoid replacing all indexes and aliases with environment specific indexes

The problem I have is if I set a custom Analyzer, this won't be picked up. This results in a MapperParsingException: Analyzer [sortable] not found for field [raw] 

The only way round I have is an ugly hack of creating an empty index called 'index_v1' that does nothing other than set the Settings. 

Is this an issue, and has it or will it be addressed in later versions?

Thanks

eilgin:
This should fix #853 when a user reset a specific type in an index and wants his changed settings (_analyzer_, _filters_, etc) to be taken into account.

### DIFF
--- a/Index/Resetter.php
+++ b/Index/Resetter.php
@@ -136,7 +136,9 @@ class Resetter
         }
 
         if (!empty($settings)) {
+            $index->close();
             $index->setSettings($settings);
+            $index->open();
         }
 
         $mapping = new Mapping();
--- a/Index/Resetter.php
+++ b/Index/Resetter.php
@@ -118,7 +118,11 @@ class Resetter
     public function resetIndexType($indexName, $typeName)
     {
         $typeConfig = $this->configManager->getTypeConfiguration($indexName, $typeName);
-        $type = $this->indexManager->getIndex($indexName)->getType($typeName);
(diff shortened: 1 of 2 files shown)


XWB:
Thanks.

Tobion:
This is so buggy because of #1103 and more reasons. #1103 is also only compatible with ES 2 and not ES 1 because the settings have a different structure there. And this change has no error handling at all. So when changing the settings fails, the index says in an closed state. I mean cmon.

Tobion:
To make it useable again: #1164

PR #913, opened 2015-07-23
updatable callbacks

tomfun:
In few cases appear urge to decrease cpu load on some minor entity updating.
In this pull request I've made callback system like indexable to determine is realy need to update elastica index or it's just changed some backend data fields that don't realy included in elastica index.


XWB:
Closing this outdated PR.

PR #918, opened 2015-08-06
Don't use parent identifier for objects/nested documents

### LINKED ISSUE #357: Custom identifier on nested object
As of now i can be able to specify custom identifier of the main document, but when it comes to nested objects where the nested object has its own custom identifier, populating index breaks. It looks for the specified identifier in the nested object too. This happens here, while it's recursively getting called

```php
    public function transform($object, array $fields)
    {
        $identifier = $this->propertyAccessor->getValue($object, $this->options['identifier']);
        $document = new Document($identifier);

```

How can this be solved?
Any help will be much appreciated

jvandesande:
This fixes #357

### DIFF
--- a/Tests/Transformer/ModelToElasticaAutoTransformerTest.php
+++ b/Tests/Transformer/ModelToElasticaAutoTransformerTest.php
@@ -122,6 +122,19 @@ class POPO
     {
         return $this->getUpper();
     }
+
+    public function getObjWithoutIdentifier()
+    {
+        return (object) array('foo' => 'foo', 'bar' => 'foo');
+    }
+
+    public function getSubWithoutIdentifier()
+    {
+        return array(
+            (object) array('foo' => 'foo', 'bar' => 'foo'),
+            (object) array('foo' => 'bar', 'bar' => 'bar'),
+        );
+    }
 }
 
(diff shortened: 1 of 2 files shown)


jvandesande:
@merk @stof @XWB Is there someone who can review this PR?

gnoesiboe:
+1

djbfmeel:
+1

evertharmeling:
+1

Meuk:
+1

merk:
There are no test cases to verify this fix. I need a failing test case on the original code to merge this PR.

jvandesande:
@merk I've added the requested test cases

XWB:
@jvandesande Can you fix this in 3.0 as well?

jvandesande:
@XWB I can do that. When will this PR be checked/merged?

jvandesande:
@XWB I've created a PR for the 3.0.x branch too: https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/935

stof:
@XWB I suggest we stop maintaining multiple branches in parallel. It adds more work and should not be necessary if we really follow semver (see Twig for instance: it has 2 branches only because of the next major version being in progress)

XWB:
@stof :+1: I'll prepare a final 3.0 release.

PR #924, opened 2015-08-19
Fix the progress logging for the last incomplete slice

stof:
This avoids increasing the max of the progress bar to the upper multiplier of the batch size.

@XWB @merk I made this change based on the master branch, as I suggest we stop maintaining 3 parallel branches of the bundle (3.0, 3.1 and master). We should try to maintain only the master branch (until we start working on the next major version in which case we will need a 3.x branch). This will simplify the bundle maintenance.


merk:
I agree, though right now I'm not working on any Symfony applications so my time for maintenance here is low. Once we release a 3.2 there is no point maintaining the 3.0 branch.

Issue #925, opened 2015-08-24
Unable to save the index when creating new entity


zanardigit:
Sorry for a newbie question, but I get the following exception when trying to save a new entity and I have no idea how to fix it:

    IllegalArgumentException[Malformed action/metadata line [1], expected a simple value for field [_id] but found [START_OBJECT]] [] []

The `populate` command works fine. Updating an existing entity works fine. The exception is correct, because this is an extract of the JSON passed to ES:

    {"path":"_bulk","method":"PUT","data":"{\"index\":{\"_index\":\"app\",\"_type\":\"user\",\"_id\":{}}}...

and you can see `_id` is an empty object, while on update it's complete:

    {"path":"_bulk","method":"PUT","data":"{\"update\":{\"_index\":\"app\",\"_type\":\"user\",\"_id\":\"0ac35136-7109-4b45-91c6-7bf6dd7b6bc6\"}}\n

The only difference is that I am using UUID instead if autoincrement id on this app. 
I var_dumped the entity before persist and it's complete, i.e. the `id` field is filled correctly on a new entity.

Is there any special configuration I need to do while using UUID?

zanardigit:
I think I found a solution. Shortly: the `getId` method in my entity casts the UUID to string.

# Details

After digging through FosElasticaBundle code, I see that the ID is retrieved using the standard property accessor component. And of course the property accessor component just ask the entity for it. The entity method is like this:

    public function getId()
    {
        return $this->id;
    }

and of course it returns an UUID **object**. But usually when we ask for the id, what we really want is the string and not the UUID object, so I changed the method to this:

    public function getId()
    {
        return (string)$this->id;
    }

To make FosElasticaBundle more compatible out-of-the-box in such situation, we could change a line in the method `ModelToElasticaAutoTransformer::transform` from:

    $identifier = $this->propertyAccessor->getValue($object, $this->options['identifier']);

to

    $identifier = (string)$this->propertyAccessor->getValue($object, $this->options['identifier']);

I will issue a PR and then you will see if it makes sense to merge. Thank you for this bundle!

Issam7:
Hello

I have the same error actually,
`IllegalArgumentException[Malformed action/metadata line [1], expected a simple value for field [_id] but found [START_OBJECT]]`

Any news about this ?

zanardigit:
@Issam7 if you read my comment and the comments in the attached PR, you should be able to fix it. 
You have a few options.

[...]

PR #930, opened 2015-09-03
Cast identifier to string

zanardigit:
When identifier is an object (i.e. UUID) we cast it to string when building the Elastic document. See #925 for details.

### DIFF
--- a/Transformer/ModelToElasticaAutoTransformer.php
+++ b/Transformer/ModelToElasticaAutoTransformer.php
@@ -67,7 +67,7 @@ class ModelToElasticaAutoTransformer implements ModelToElasticaTransformerInterf
      **/
     public function transform($object, array $fields)
     {
-        $identifier = $this->propertyAccessor->getValue($object, $this->options['identifier']);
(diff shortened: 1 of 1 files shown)


XWB:
What happens if the object does not have a `__toString` method?

stloyd:
Fatal error: `Catchable fatal error: Object of class XXX could not be converted to string`.

zanardigit:
@XWB @stloyd that's exactly my point. In this way you *see* that there's something wrong. AFAIK the identifier is used as a string anyway in the JSON document sent to Elastic Search, so it ends up empty if it has not a __toString method.

merk:
Use the property path feature of the mapping to point to a getter that returns the string wish to use for the identifier.

zanardigit:
@merk thank you for your reply. 
I don't need to change my mapping. I only have to make sure my object has a `__toString` method. 
My PR was to avoid any pre-requisite and automatically convert any given identifier to string, and **show** a clear error if it can not. It took a while for me to find out this pre-requisite.
Anyway, you're the maintainer so it's up to you :-) 
Thanks everyone for their feedbacks.

PR #942, opened 2015-09-17
Add fos:elastica:reindex command.

aheinz-sg:
Ok! This is a work in progress, but seems to be working well in initial manual testing. I've got plenty of work left to do, but constructive criticism is welcome. Known issues:
1. Delete unnecessary options copy/pasted from fos:elastica:populate
2. Use logger closure instead of output.writeln
3. Tests
4. Dispatch reindex events?

### DIFF
--- a/Command/ReindexCommand.php
+++ b/Command/ReindexCommand.php
@@ -150,6 +150,7 @@ class ReindexCommand extends ContainerAwareCommand
             $resultSet = $scanAndScroll->current();
             $resultCount = $resultSet->count();
 
+            // Process the current batch
             try {
                 $type->addObjects($resultSet->getResults());
             } catch (\Exception $e) {
--- a/Command/ReindexCommand.php
+++ b/Command/ReindexCommand.php
@@ -1,0 +1,190 @@
+<?php
+
+namespace FOS\ElasticaBundle\Command;
+
+use Elastica\Query;
+use Elastica\Search;
+use Elastica\Type;
+use FOS\ElasticaBundle\Elastica\Index;
+use FOS\ElasticaBundle\Event\IndexPopulateEvent;
+use FOS\ElasticaBundle\Event\TypePopulateEvent;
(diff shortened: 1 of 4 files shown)


aheinz-sg:
I guess I should describe what this command does! It creates a new index, then bypasses the Symfony ORM provider entirely and does Elasticsearch index -> index reindexing, then cuts over the alias. Some of the reading that inspired this feature:  
1. https://www.elastic.co/guide/en/elasticsearch/guide/current/reindex.html
2. https://www.elastic.co/blog/changing-mapping-with-zero-downtime#_reindexing_your_data_with_zero_downtime

aheinz-sg:
Ok, looks good! Going to put this onto my development cluster today and smash millions of documents through it to see how it performs.

aheinz-sg:
Tests at scale are looking good! I just jammed around 8M documents through fos:elastica:reindex in a little more than an hour with a peak memory usage of 256MB. Still single threaded for now, but I suspect a scroll id can be shared across processes... an investigation for a future PR.

IMPORTANT: The command will exhaust memory if...
1. You set monolog.handlers.main.type: fingers_crossed
2. You fail to set logger: false on your fos_elastica.clients.default.connections

In both cases, the logger accumulates requests/responses in memory... all the requests/responses... *facepalm*

Command/ReindexCommand.php
```diff
@@ -0,0 +1,202 @@
+<?php
+
+namespace FOS\ElasticaBundle\Command;
+
+use Elastica\Search;
+use Elastica\Type;
+use FOS\ElasticaBundle\Elastica\Index;
+// use FOS\ElasticaBundle\Event\IndexPopulateEvent;
+// use FOS\ElasticaBundle\Event\TypePopulateEvent;
```
piotrantosik: delete line with comments

damienalexandre:
Thanks for this feature :+1: I like it!

There is some cleaning to be done on the code and documentation to write, because the command has a lot of options.

ggiovinazzo:
Any news on this feature?

PR #945, opened 2015-09-28
Add PaginatorAdapterInterface as one of FantaPaginatorAdapter's interfaces

lemoinem:
Not having `PaginatorAdapterInterface` as one of `FantaPaginatorAdapter`'s interfaces kind of go against the purpose of having a `PaginatorAdapterInterface` in the first place...

### DIFF
--- a/Paginator/FantaPaginatorAdapter.php
+++ b/Paginator/FantaPaginatorAdapter.php
@@ -4,7 +4,7 @@ namespace FOS\ElasticaBundle\Paginator;
 
 use Pagerfanta\Adapter\AdapterInterface;
 
-class FantaPaginatorAdapter implements AdapterInterface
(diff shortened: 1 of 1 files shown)


lemoinem:
Scrutinizer build says canceled. So I'm guessing it's related to the project's configuration and not to my modifications.

XWB:
Thanks.

ruudvanderweijde:
This results in a fatal error: https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1040

lemoinem:
@steffenbrem Well, `PaginatorAdapterInterface` seems to be the base interface for the returned value from `FINDER->findPaginated()->getAdapter()`, which seems to be the only way to have both paginated results and access to the aggregations.

Did I miss something? The names of these interfaces and how they relate to each other seems a bit confusing...

PR #946, opened 2015-10-01
Allow ordering by unmapped fields

dariuszz123:
Prevent error caused by ordering on empty fields


pich:
Darius you saved my day.

XWB:
Thanks.

PR #947, opened 2015-10-01
Made default sort options from pagination config to work as expected

dariuszz123:
Fallback to defaultSortFieldName and defaultSortDirection pagination options instead of ignoring then.


XWB:
Any idea why Scrutinizer fails?

dariuszz123:
No idea, strange status "Canceled", I think it was some kind of Scrutinizer system internal error... Any ideas how to rerun Scrutinizer tests?

XWB:
@dariuszz123 I would rename the branch and open a new PR.

PR #953, opened 2015-10-27
Allow driverless types

merk:
This change allows the definition of a type with its own custom persister implementation without needing to define a driver type.


merk:
Note: I've made this change because I have types that are coming in from non persistent backends that I want FOSElasticaBundle to handle. It isnt ideal: i'd prefer to rewrite the way we handle defining indexes and types, but for now this will have to do because a rewrite is a big job.

PR #954, opened 2015-11-03
Add dynamic option in mapping at the type and field level

damienalexandre:
This PR add support for the `dynamic` option in mapping.

https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html

This option allow to control when Elasticsearch is allowed to dynamically map an unknown field. I added as an example what is known to be a good practice, which is disabling the dynamic mapping completely, and then allowing it for some particular fields.


damienalexandre:
It's been a month, so I would like to know, is there anything I can do to help?

This is a documented Elasticsearch feature, and my PR include documentation, tests and example.

Thank you! :beers:

PR #958, opened 2015-11-11
es 1.7 function_score is incomplete

Raffaello:
field value factor is missing in elastica until 2.3.1

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -434,7 +434,6 @@ class Configuration implements ConfigurationInterface
             ->scalarNode('enabled')->defaultValue(true)->end()
             ->scalarNode('index_analyzer')->end()
             ->scalarNode('search_analyzer')->end()
-            ->scalarNode('analyzer')->end()
             ->end()
         ;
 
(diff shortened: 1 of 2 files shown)


Raffaello:
anyone on this?

merk:
There doesnt appear to be any relevant changes for this PR?

Raffaello:
in the composer.json is not using the ruflin/elastica dependency that have the some functionality such as:
function_score query etc...

-        "ruflin/elastica": "~2.1,<2.3",
+        "ruflin/elastica": "~2.1,<2.3.*",

Raffaello:
it means it is not possible to use with foselasticabundle this function:

https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-function-score-query.html

with some other parameters.

XWB:
This is not the right way to resolve an issue, we should make it work on 2.3.1 too.

Raffaello:
exactly, the issue is just the composer.json constrain on ruflin/elastica
that dsl are implemented in the 2.3.1 and actually fos/elastica bundle strictly express to use less than 2.3 of ruflin/elastica
@XWB the 2.3.1 is rufflin/elastica dependency not FOSElasticaBundle.
there is no reason why not to upgrade the ruflin/elastica in FOSElastica

PR #962, opened 2015-11-17
Update Provider.php

Raffaello:
removed the optional groupBy clause from the count query

if it is provided a custom query builder method with a groupBy clause the count method will fail.

### DIFF
--- a/Doctrine/ORM/Provider.php
+++ b/Doctrine/ORM/Provider.php
@@ -64,6 +64,7 @@ class Provider extends AbstractProvider
             ->select($qb->expr()->count($rootAliases[0]))
             // Remove ordering for efficiency; it doesn't affect the count
             ->resetDQLPart('orderBy')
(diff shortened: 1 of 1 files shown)


Raffaello:
is it possible to merge it?

merk:
A group by could fundamentally change the query being run and the count returned?

Please provide an example for the query you want to run where removing the group by has no effect. You can work around this by defining another custom method that doesnt provide a QB with the group by.

Raffaello:
ok. 
Honestly I don't remember exactly, but it should be: 
It was caused by using some specific Mysql custom SQL for the custom querybuilder where to fetch the data. It was using the "groupConcat" function with a Doctrine extension bundle.

the other iterations were returning empty results, or better, a wrong query because it was partially "messed up"  with the first iteration.

XWB:
I agree with @merk 

I'm closing this PR, feel free to open a new issue if you encounter the problem again.

PR #965, opened 2015-11-17
Allow set retryOnConflict per connection

### LINKED ISSUE #807: RetryOnConflict
Hello,
I have problems with my index, after a populate everything is right, but in times some elements start missing, even worst: my website start to throw errors like this one:
```
[2015-03-03 23:46:36] request.CRITICAL: 
Uncaught PHP Exception Elastica\Exception\Bulk\ResponseException:
"Error in one or more bulk request actions: update: /app/type/2 caused
VersionConflictEngineException[[index][3] [type][2]: version conflict, current [752], provided [751]] "
at /var/www/App/vendor/ruflin/elastica/lib/Elastica/Bulk.php line 395
```

I found out the error was caused by a double update. But I found I could fix it by a force update thanks to "retryOnConflict". (https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Client.php#L24)

Do I have a way to set it in my config, does someone already used/set this ? Do you have another solution to my problem ?

Thanks a lot for your help !

piotrantosik:
Fix #807


merk:
Thanks

PR #966, opened 2015-11-20
Update types doc page

dymissy:
The PR wants to update documentation in order to include new PHPCR driver supported.

As @dbu suggested, I just updated the `types` page.


dbu:
thanks. this already gives a hint that its possible, so i merge this. further additions are welcome, though we should not make that documentation too large.

PR #967, opened 2015-11-24
Added methods to get suggests out of a paginated result

likeitlikeit:
It seems like these methods would come real handy when using KnpPaginator together with FOSElastica. When the search query includes suggests, it is otherwise impossible to get them from the processed query. If this should go somewhere else, please let me know.


likeitlikeit:
Fixes #968.

merk:
Please add test coverage.

XWB:
We are not maintaining 3.1 anymore.

PR #970, opened 2015-11-27
Differentiate query time from execution time

bastnic:
The only time showed by this bundle is the whole time taken by the request : latency + query + fetch. So a fast query but a long fetch is marked as slow in the debug toolbar. 

With this patch I want to address this case : 
- query 1: 100ms executionTime 
- query 2 : 20ms executionTime

Naturally I want to optimize query 1, but with more analysis, what happened is that : 
- query 1: 100ms executionTime, 1ms queryTime
- query 2 : 20ms executionTime, 15ms queryTime

So the fix is different : I need to work on the query itself on query 2 and on the data retrieved on query 1.

I also return the number of result given by the query, as it can give a useful info about the query (hello the sort on 1000000 results, better shrink the resultset).

This is a very naive approach and we sure can add more time information (latency?) to have some useful time given. 

What do you think?

### DIFF
--- a/DataCollector/ElasticaDataCollector.php
+++ b/DataCollector/ElasticaDataCollector.php
@@ -44,7 +44,7 @@ class ElasticaDataCollector extends DataCollector
     {
         $time = 0;
         foreach ($this->data['queries'] as $query) {
-            $time += $query['executionMS'];
+            $time += $query['engineMS'];
         }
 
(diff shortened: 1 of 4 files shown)


delphiki:
Could be really useful indeed :+1:

damienalexandre:
:+1:

PR #971, opened 2015-11-27
Allow serialize_null option for Serializer

michaelperrin:
It is currently impossible to populate attributes with a `null` value with FOSElasticaBundle, due to to the default behavior of JMS Serializer which doesn't serialize attributes with `null` values.

This PR adds an option to the `serializer` configuration so that the JMS Serializer can be configured with the `serialize_null` option.

There is no specific test for this PR, as it is a configuration for JMS Serializer, and I didn't find any other test concerning its configuration (for the `version` and `groups` options). I just updated the default test configuration (with a `false` value, which is the default value anyway), and added an example to the documentation for better clarity.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -587,6 +587,9 @@ class Configuration implements ConfigurationInterface
                     ->prototype('scalar')->end()
                 ->end()
                 ->scalarNode('version')->end()
+                ->booleanNode('serialize_null')
+                    ->defaultFalse()
+                ->end()
             ->end();
 
         return $node;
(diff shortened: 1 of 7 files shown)


DependencyInjection/Configuration.php
```diff
@@ -583,6 +583,8 @@ protected function getSerializerNode()
                     ->prototype('scalar')->end()
                 ->end()
                 ->scalarNode('version')->end()
+                ->booleanNode('serialize_null')
+                    ->defaultFalse()
```
stloyd: Missing `end()` method call.

michaelperrin: That was fast @stloyd, thanks!
It is fixed now.

merk:
Please add test coverage for this change

michaelperrin:
Hi @merk, I added the needed tests.

I make sure that `null` values are not serialized (and therefore not persisted) when the `serialize_null` option is set to `false` (which is the default and current behavior), and then make sure it is defined and equals to `null` when persisted with a `true` value for the option.

michaelperrin:
Hello, any news on this PR please?

XWB:
Thanks.

michaelperrin:
Thanks @XWB !

PR #972, opened 2015-11-30
Don't add $property['store'] = true for type nested and attachment

FireFoxIXI:
Adding 'store': true to those types makes the mapping invalid in elasticsearch 2.0


damienalexandre:
:+1:

merk:
This isnt necessary, I'll be committing a change shortly that removes the automatic setting of store:true unless it is specifically specified in the mapping.

PR #977, opened 2015-12-04
Use single option in setAllowedTypes

xurshid29:
Will be compatible  with the 3rd version of symfony .

### DIFF
--- a/Provider/AbstractProvider.php
+++ b/Provider/AbstractProvider.php
@@ -108,9 +108,7 @@ abstract class AbstractProvider implements ProviderInterface
             'batch_size' => 100,
             'skip_indexable_check' => false,
         ));
-        $this->resolver->setAllowedTypes(array(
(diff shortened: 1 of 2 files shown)


stloyd:
This is wrong, you should call `method_exits()` for new method, otherwise it's broken for Symfony `2.3`.

xurshid29:
Ah, I see.

xurshid29:
@stloyd I'll fix it.

xurshid29:
@stloyd I don't think `method_exits()` can help here. `setAllowedTypes` in 2.3 version uses type hinting, which doesn't allow to pass a string.

xurshid29:
I pushed the commit which fixes the 2.3 compatibility. The solution is not so pretty, but I couldn't find a better way.

merk:
I removed the entire setAllowedTypes section to simplify this.

Issue #985, opened 2015-12-20
Mapping definition has unsupported parameters:  [store : true]


mappedinn:
I defined my index `ads` as follows:

    fos_elastica:
        clients:
             default: { host: %elastica_host%, port: %elastica_port% }         
        indexes:
              ads:
                types:                        
                    brand:
                        mappings:
                            name:
                                type: string
                                boost: 5
                            slug:
                                type: string
                                boost: 4
                            date : ~
                            isPublished: ~
                            user:
                                type: nested
                                properties: 
                                    username: ~
                                    email: ~
                        persistence:
                            driver: orm
                            model: Minn\AdsBundle\Entity\Brend
                            elastica_to_model_transformer:
                                service: minn_ads.transformers_elastica.brand
                            provider: ~
                            listener:
                                immediate: ~
                            finder: ~

**FYI:** 1. This is how `Brend` is linked with `User` with @ManyToOne

    /**
     * @ORM\ManyToOne(targetEntity="Minn\UserBundle\Entity\User")
     * @ORM\JoinColumn(nullable=false)
     */
    private $user;

**FYI:** 2. I am using the `dev-master` branches for  FOSElasticaBundle and Elastica. For elasticsearch, I am using 2.1.1.

A populate command `php app/console fos:elastica:populate` always returns this error:

 **[Elastica\Exception\ResponseException]                                                                                                                                                                                 
  {"root_cause":[{"type":"mapper_parsing_exception","reason":"Mapping definition for [user] has unsupported parameters:  [store : true]"}],"type":"mapper_parsing_exception","reason":"Failed to parse mapping [brand]:  
   Mapping definition for [user] has unsupported parameters:  [store : true]","caused_by":{"type":"mapper_parsing_exception","reason":"Mapping definition for [user] has unsupported parameters:  [store : true]"}}** 

I checked the `app/logs/dev.log`, I found that the generated mapping for the index `ads` has an extra parameter `"store":true`. You can check that down:

[...]

PR #987, opened 2015-12-21
simple fixing of the bug adding the property store:true

mappedinn:
Hi,

This PR fixes the bug of the issue #985  that adds for each property `"store": true` by simply commenting the code in charge of it in `Index/MappingBuilder.php`.

### DIFF
--- a/Index/MappingBuilder.php
+++ b/Index/MappingBuilder.php
@@ -21,7 +21,7 @@ class MappingBuilder
      *
      * @var array
      */
-    private $skipTypes = array('completion');
+    private $skipTypes = array('completion','nested');
 
     /**
      * Builds mappings for an entire index.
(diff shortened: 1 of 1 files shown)


merk:
I'll be pushing a commit shortly that removes setting store:true by default, which negates the need for this change.

ngeleedst:
Is their any news about this fix. Because now i cannot populate because it's adding store: true to a type: object etc.. and this is not allowed!

            if (!isset($property['store']) && $property == 'string') {
                $property['store'] = true;
            }

ngeleedst:
Or wait maybe just by adding:  $skipTypes = array('completion','nested','object'); 
Shall i do a pull request with the new type 'object' added to the skypTypes?

ngeleedst:
I did a new pull request regarding this ticket. See: [https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1013](url)

PR #988, opened 2015-12-21
Store fixed

mappedinn:
Hi, This PR fixes the bug of the issue #985  that adds for each property `"store": true`.

### DIFF
--- a/Index/MappingBuilder.php
+++ b/Index/MappingBuilder.php
@@ -21,7 +21,7 @@ class MappingBuilder
      *
      * @var array
      */
-    private $skipTypes = array('completion');
+    private $skipTypes = array('completion','nested');
 
     /**
      * Builds mappings for an entire index.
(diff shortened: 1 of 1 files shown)


mappedinn:
@merk, @richardmiller: can you have a look on this small fix?

foaly-nr1:
+1

foaly-nr1:
@mappedinn Reading @merk's comment on your initial bug report, I think he envisaged something different from this PR. I think the status property shouldn't automatically be added at all (because within ElasticSearch it defaults to true) and tests need to be added.

> Send a PR that removes store:true from MappingBuilder (since it is default it shouldnt be there unless false), provided there are added tests for the change I'm happy to merge it

mappedinn:
@foaly-nr1 you are right. But, I proposed this PR as **quick fix** because the bundle does not work unless you correct MappingBuilder.

merk:
Quick fixes do not get merged without tests.

I'll be pushing a commit shortly that removes setting store:true by default, which negates the need for this change.

mappedinn:
thank you very much @merk

PR #993, opened 2015-12-28
Symfony3 compatibility

vworldat:
Fixed code and tests to get full 2.8/3.0 compatibility (as far as the current test coverage goes). This results in loss of Symfony 2.3 support - the minimum version is now 2.6.

I don't know if this is desired at the moment, but on the long run it will be necessary anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,13 +15,13 @@ php:
 matrix:
   include:
     - php: 5.6
-      env: SYMFONY_VERSION='2.3.*'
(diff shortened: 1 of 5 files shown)


Jupdi:
Great! Done almost the same fixes locally to get the bundle running with 3.0!

merk:
Symfony 2.3 support must remain until it reaches EOL.

merk:
Pushed a commit that should fix 3.0 compatibility

merk:
(as part of 3.2.0@dev)

PR #1001, opened 2016-01-12
Show query backtrace in profiler

ewgRa:
Often there is needed to understand where ES query come from. This PR add backtrace information and show it in profiler. It is easy to find line, that raise query call:

profiler-query-backtrace


XWB:
Thanks.

Issue #986, opened 2015-12-21
Roadmap


merk:
I havent had much time lately to do any work on this bundle, but I have spent a lot of downtime thinking about the issues that keep coming up, the problems with the current implementation and where the bundle needs to go.

I'm hoping this information being out there will inform or encourage contributors to help out where they can, since the bundle is currently in a state that is suitable for any of my projects and their limited search requirements.
## Developer experience and expectations

I'm seeing a lot of issues created that are just questions about using ElasticSearch itself, or specific questions for Elastica. We need to make it clearer, by removing code and options in FOSElasticaBundle that the bundle is for integration with Elastica, we dont replace it.

The biggest issue is the paginators and the way the bundle encourages searching - you need to use FOSElastica specific interfaces to search and get transformed results and it sucks.

What I want to see is the removal of any pagination support and finder services from the bundle. Pagerfanta and KnpPaginator both support paginating Elastica queries, and with examples that clearly show standard use of Elastica, we can reduce confusion here (hopefully!)
### Transformations

Since Elastica doesnt support transformations out of the box, FOSElastica has had to implement transformations (both ways), and I've felt that this code doesnt belong here anymore, it belongs in Elastica or an Elastica-Extras repository that handles both serialization of objects to ElasticSearch representations and retrieval of persisted objects based on search results and providing that object through the Elastica\Result object.
### Mappings

Managing mappings in the DIC is clunky, bug ridden and sucks. I dont have a solution to this one, mapping management could also belong in Elastica, since anyone using it could benefit from a programmatic interface for management of mappings.

The mapping metadata also ties in with the Transformations topic above, since knowledge of mappings helps transforming objects appropriately.

merk:
A short tldr version:

- Move mapping management to an external reusable library, or integrate into Elastica
- Move transformations (both Elastica->Model and Model->Elastica) into Elastica or another external library
- Strip back FOSElasticaBundle to a minimal surface area integrating these pieces, exposing raw Elastica services for use in applications.

merk:
Expanding on the mappings issue, I havent been able to come up with a great solution, the 3 i've considered:

- *Fixing the DIC based configuration*: a non solution, it becomes a nightmare to keep the options in sync with options available in different ElasticSearch versions
- *A annotation/yml/xml based approach like Doctrine's metadata*: A decent approach, but with the same limitations of different options for different ES versions
[...]

PR #1002, opened 2016-01-13
Bring the API to match Elastica 3.0.1

emgiezet:
Hi!,

I made most of the changes. But still one test don't pass. Can anybody help me with the:  MappingToElasticaTest:testResetIndexAddsMappings

The Bundle work with 14 GB Index (production dump tested) But still test fails.

Regards!
Max


kamdz:
:+1:

kamillo1888:
:+1:

sebastiancichos:
+1 we need this too...

emgiezet:
Any luck with this?

dunglas:
I confirm that it works for me too.

Tests/Functional/MappingToElasticaTest.php
```diff
@@ -21,11 +21,13 @@ class MappingToElasticaTest extends WebTestCase
     public function testResetIndexAddsMappings()
     {
         $client = $this->createClient(array('test_case' => 'Basic'));
-        $resetter = $this->getResetter($client);
-        $resetter->resetIndex('index');
 
         $type = $this->getType($client);
         $mapping = $type->getMapping();
+        $resetter = $this->getResetter($client);
```
pascaldevink: It seems that by juggling the location of the resetter, ```$this->getType($client)``` is using an index that hasn't been created yet. Resetting this to the original position fixes the current test error.

pascaldevink: However, a lot of other things have been deprecated in ES2.x, so other errors will come up after that.

pascaldevink: I've done a PR against your PR (PR-inception!) to address these issues: https://github.com/emgiezet/FOSElasticaBundle/pull/1

DependencyInjection/Configuration.php
```diff
@@ -417,8 +417,8 @@ protected function getParentNode()
         $node
             ->children()
                 ->scalarNode('type')->end()
-                ->scalarNode('property')->defaultValue(null)->end()
-                ->scalarNode('identifier')->defaultValue('id')->end()
+//                ->scalarNode('property')->defaultValue(null)->end()
+//                ->scalarNode('identifier')->defaultValue('id')->end()
```
XWB: This should be removed.

DependencyInjection/Configuration.php
```diff
@@ -417,8 +417,8 @@ protected function getParentNode()
         $node
             ->children()
                 ->scalarNode('type')->end()
-                ->scalarNode('property')->defaultValue(null)->end()
-                ->scalarNode('identifier')->defaultValue('id')->end()
+//                ->scalarNode('property')->defaultValue(null)->end()
```
XWB: This should be removed.

Paginator/RawPartialResults.php
```diff
@@ -40,13 +40,10 @@ public function getTotalHits()
 
     /**
      * {@inheritDoc}
+     * @deprecated remvem from rufflin api 3.0.1
```
XWB: Typo?

composer.json
```diff
@@ -12,12 +12,12 @@
         { "name": "Jeremy Mikola", "email": "jmikola@gmail.com" }
     ],
     "require": {
-        "php": ">=5.3.2",
+        "php": ">=5.4.0",
```
XWB: Why would we bump PHP to 5.4?

composer.json
```diff
@@ -12,12 +12,12 @@
...
         "symfony/framework-bundle": "~2.3|~3.0",
         "symfony/console": "~2.1|~3.0",
         "symfony/form": "~2.1|~3.0",
         "symfony/property-access": "~2.3|~3.0",
-        "ruflin/elastica": "~2.1",
+        "ruflin/elastica": "3.0.*",
```
XWB: Not sure if we want to drop Elastic 1.7 support, ping @merk

.travis.yml
```diff
@@ -5,7 +5,6 @@ cache:
         - $HOME/.composer/cache
 
 php:
-  - 5.3
```
XWB: What exactly goes wrong on PHP 5.3?

emgiezet: The Es 3.0.1 have 5.4 in dependencies

XWB: I see.

Just noticed PHP 5.3 is EOL http://php.net/supported-versions.php, I suppose now is a good time to drop 5.3 support.

emgiezet:
i've rebased pr to newest master.

merk:
Not willing to drop Elastica 2.x support or PHP 5.3 support until we bump major versions of FOSEB.

emgiezet:
@merk this is sad :(

dunglas:
@merk Support for Elastic 3 can maybe be added to another branch?

merk:
The version required for using this bundle should not be locked down by what version of Elastica we want to support. The fact that Elastica has hard dependencies on specific ES versions means that we're locking out users who cannot upgrade from ES1.7.

This is not acceptable. I'm happy to merge any code that maintains BC. A separate branch is a no go, we do not have enough contributors or time allocated to management of this bundle to maintain 2 branches.

emgiezet:
I'm using foselastica with a huge set of data. We implemented own populate command to gain multithreaded performance.
We skipped the listener transformer and put data to the ES server by async http request with guzzle and batch requests.

But still I can't see the logic after the decision to not follow version of your main and only dependant package.

I'm using elasticsearch since 0.6 beta. The changelog from 1.7 compared with current version is huge (ex. populating straight from mongo).

In this case I wish you guys luck. I must use pure rufflin/elastica instead,

And when you wondering why people don't contribute much. The answer is simple. 

> Because they don't want to waste their work.

Cheers!
Max

pascaldevink:
@merk So that means that nobody can upgrade their ES cluster beyond 1.7.x if using this bundle? That would be a great way to loose almost everybody that is currently using it and not gaining any new users, as they would start out with the current ES version (2.2.x).

Wouldn't it be better to start working on a new major version with support for newer ES version? If people still want to use older versions of ES, and this bundle, that's fine, but let's try to cater more to the people that use the latest versions instead of trying to keep backwards compatibility for the people that refuse to upgrade (for whatever reason).

PHP 5.3 support was dropped over a year ago and ES 5 is already announced.

merk:
So should we abandon users who don't have the capability to upgrade every time a new version of a product comes out? Symfony 2.3 is still supported and runs on PHP 5.3.

The truth is we use none of the new features of either Elastica or ElasticSearch and it should not be such a pain to support multiple versions. I've yet to see anything specific that cant be handled without breaking BC.

Theres a major problem: the bundle is doing too much in a way that just doesn't make it very useful for anything except a single case. Do you continue to develop this code base, which is sorely missing tests, or look at a new approach? I've not see many people comment when I've tried to raise the topic, nor offer to assist in making the situation better.

My contribution to Elastica to make it simpler exposes a big issue: since FOSEB doesnt actually use much in the way of Elastica's features, why dont we move to just using the official client? It already supports some of the things we need for the bundle.

https://github.com/ruflin/Elastica/pull/1017
https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/986

kwojc:
@merk @XWB Last release of PHP 5.3 was in 2014, don't be silly, we are in 2016. If somebody can't bump up version of PHP, he should use older version of FOSEB IMO, next major version of FOSEB is needed urgently, people likes to use fresh vendors with support to newest versions of software.

pawelkolanowski:
@kwojc +1

merk:
You've missed the point here, dropping 5.3 isnt necessary for this PR. The actual point is I'm not dropping BC at this time. 

I'm actually tempted to deprecate the entire bundle since there is almost nothing in the way of help coupled with a lack of any real test coverage. It might be simpler to start over.

Gladhon:
@merk 
Dropping 5.3 is necessary to update elasatic libary. If you cant update your own dependencies, you always stay on this buggy and feature-less version.

mojzis:
@merk i think i understand most of your points, but still we suffer from not being able to use some stuff in ES 2 ... how about a compromise where this would go **temporarily** into a new branch that would also be a new major release and all the cleanup and refactoring would be done in yet another major release ?
Trying to support both versions of Elastica looks like something noone is willing to contribute now and since there is the working code from @emgiezet I think it would be a pity not to use it for the good of many projects dependent on FOSEB.
Thanks for considering the compromise :).

.travis.yml
Soullivaneuh: > Just noticed PHP 5.3 is EOL http://php.net/supported-versions.php, I suppose now is a good time to drop 5.3 support.

Yes, but it's have to be done a new major.

composer.json
Soullivaneuh: `3.0.*` -> `^3.0`.

> Not sure if we want to drop Elastic 1.7 support, ping @merk

Any chance to support both version of `ruflin/elastica` to make it BC?

XWB:
Closing this in favor of https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1069

PR #1007, opened 2016-01-19
Allow passing extra parameters to queryBuilder factory

kix:
This adds a possibility to not only specify a method that creates a query builder, but also to supply extra parameters for the method call.

The previous implementation didn't really allow much flexibility when creating a query builder; just having a single method called at all times for a given type is a constraint that has limited several use cases for me, and I think I might be not the only one having this kind of trouble.

Note: this is in an un-tested state, I should fix this.


XWB:
@kix Can you rebase from master?

kix:
@XWB, rebased.

XWB:
Thanks.

PR #1008, opened 2016-01-21
[Profiler] Match 2.8 redesign

bgaleotti:
**Toolbar**
toolbar

**Toolbar detail**
toolbar-detail

**Menu profiler**
menu-profiler


Resources/views/Collector/elastica.html.twig
```diff
@@ -1,48 +1,60 @@
...
+            <span class="sf-toolbar-status">{{ collector.querycount }}</span>
+            {% if collector.querycount > 0 %}
+                <span class="sf-toolbar-info-piece-additional-detail">in {{ '%0.2f'|format(collector.time * 1000) }} ms</span>
+            {% endif %}
+        {% else %}
+            {{ include('@FOSElastica/Collector/icon.svg') }}
+            <span class="sf-toolbar-value">{{ collector.querycount }}</span>
+            {% if collector.querycount > 0 %}
```
javiereguiluz: In the new toolbar, when a panel does not provide information, we don't display it. In this case, when there are 0 queries, this panel contains no information. If you want to follow this practice (which is not mandatory) you can change this:

```twig
{% include 'WebProfilerBundle:Profiler:toolbar_item.html.twig' with { 'link': profiler_url } %}
```

by this:

```twig
{% if collector.querycount > 0 %}
    {% include 'WebProfilerBundle:Profiler:toolbar_item.html.twig' with { 'link': profiler_url } %}
{% endif %}
```

javiereguiluz:
I left a minor comment, but the pull request looks nice. Good work!

merk:
Thanks!

PR #1013, opened 2016-02-01
Skip array, object and nested file in MappingBuilder

ngeleedst:
I added the following types: array, object and nested to the skipTypes array. I've seen the pull reject from a month ago: https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/987. 
I have been waiting since that day for a fix, but this is still a issue in the current version, so thats the reason why i added them and made a new pull request.

### DIFF
--- a/Index/MappingBuilder.php
+++ b/Index/MappingBuilder.php
@@ -21,7 +21,7 @@ class MappingBuilder
      *
      * @var array
      */
-    private $skipTypes = array('completion', 'array', 'object');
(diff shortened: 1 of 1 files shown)


merk:
The fix to remove store being set was released in 3.1.8 on the 7th of January

https://github.com/FriendsOfSymfony/FOSElasticaBundle/commit/19ca605228c3baabd4e57f120528e3571c9d7647

PR #1015, opened 2016-02-04
Allow to access to the Type in transformers and events

### LINKED ISSUE #908: Why the POST_TRANSFORM event occurs before setting the document type
I need to add custom properties based on the document type. However, if i try to get the document type (`$event->getDocument()->getType()` inside the callback with `$event` as the transform event parameter), i'll get a `null` value. 
It seems that the document type is set just after in `addDocuments`. Why not setting the type (_and the index_) before dispatching the transform event ?

dunglas:
Close #908.

Propagates the current type to transformers and events. This is useful when you need to know on which index you are working on.

My use case: I store translations in a different index for each lang. I want to set some custom properties on a given lang, I need to know the name of the current index.

### DIFF
--- a/Event/TransformEvent.php
+++ b/Event/TransformEvent.php
@@ -12,6 +12,7 @@
 namespace FOS\ElasticaBundle\Event;
 
 use Symfony\Component\EventDispatcher\Event;
+use Elastica\Type;
 
 class TransformEvent extends Event
 {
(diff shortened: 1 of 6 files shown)


nicolasricci:
+1
thanks @dunglas! 
can this be reviewed to be merged?

dunglas:
ping @merk

XWB:
Closing this outdated PR; injecting the Type in Transformers does not feel right.

diimpp:
Why it is not right? I have multi-locale data set and I would like to add custom properties based on current locale (index). 

If I'm correct, it's not possible currently and I'm forced to mix different languages inside same index.

PR #1019, opened 2016-02-06
Fixed bug where sort field was requested from a Request object that m…

gnoesiboe:
…ight not have been set

Resulting in the error:

```json
{
  "error": {
    "code": 500,
    "message": "Internal Server Error",
    "exception": [
      {
        "message": "Error: Call to a member function get() on null",
        "class": "Symfony\\Component\\Debug\\Exception\\FatalErrorException",
        "trace": [
          {
            "namespace": "",
            "short_class": "",
            "class": "",
            "type": "",
            "function": "",
            "file": "/projects/freshheads-rapportage/vendor/friendsofsymfony/elastica-bundle/Subscriber/PaginateElasticaQuerySubscriber.php",
            "line": 52,
            "args": []
          },

...

```

### DIFF
--- a/Resources/config/config.xml
+++ b/Resources/config/config.xml
@@ -35,6 +35,9 @@
             <call method="setRequest">
                 <argument type="service" id="request" on-invalid="null" strict="false" />
             </call>
+            <call method="setRequestStack">
+                <argument type="service" id="request_stack" on-invalid="null" strict="false" />
+            </call>
             <tag name="knp_paginator.subscriber" />
         </service>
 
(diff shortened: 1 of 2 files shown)


jvandesande:
I think that this is caused by the "request" service that has been removed in Symfony 3.0. If the "request" service exists, the request service is injected in this subscriber. See:
https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/197e90539977b1ae8fd6f4d31bb05940bd217829/Resources/config/config.xml#L3

jvandesande:
+1

rybakit:
I've created a similar PR which doesn't rely on deprecated `request` service: https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1021.

gnoesiboe:
@rybakit — ok, that's pretty much how I setup things the first time, but I changed things afterwards to keep things Symfony 2.3 compattible. (kept the `setRequest`)

stloyd:
IMO this is over complicated as you shouldn't care about whole request stack, so adding it is overkill.

Better approach would adding compiler pass that if request stack service exists would inject master request directly to this subscriber, otherwise should inject old good request.

XWB:
Closing in favor of https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1031

PR #1021, opened 2016-02-08
Fix "fos_elastica.paginator.subscriber" service instantiation

rybakit:
The `PaginateElasticaQuerySubscriber` object is not properly initialized on symfony > v3.0 as it use `request` service which was deprecated in v2.4 and removed in v3.0. 

This PR rises minimal `framework-bundle` version to 2.4 to avoid clutter with checking for existence of both `request` and `request_stack` services.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ php:
 matrix:
   include:
     - php: 5.6
-      env: SYMFONY_VERSION='2.3.*'
+      env: SYMFONY_VERSION='2.4.*'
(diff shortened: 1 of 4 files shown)


bertoost:
:+1: Please merge.

XWB:
IMO we should maintain Symfony 2.3 support.

XWB:
Closing in favor of https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1031

PR #1017, opened 2016-02-05
Fixed inject of invalid service request

desarrolla2:

PR #1031, opened 2016-03-01
Make code to be compatible with Symfony 2.4+

stloyd:
Replaces #1017, #1019, #1021

### DIFF
--- a/Subscriber/PaginateElasticaQuerySubscriber.php
+++ b/Subscriber/PaginateElasticaQuerySubscriber.php
@@ -2,19 +2,30 @@
 
 namespace FOS\ElasticaBundle\Subscriber;
 
-use Symfony\Component\HttpFoundation\Request;
-use Symfony\Component\EventDispatcher\EventSubscriberInterface;
-use Knp\Component\Pager\Event\ItemsEvent;
 use FOS\ElasticaBundle\Paginator\PaginatorAdapterInterface;
 use FOS\ElasticaBundle\Paginator\PartialResultsInterface;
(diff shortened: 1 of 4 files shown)


XWB:
Looks good.

Subscriber/PaginateElasticaQuerySubscriber.php
```diff
@@ -2,19 +2,30 @@
...
+    /**
+     * @param RequestStack|Request $requestStack
+     */
+    public function setRequest($requestStack)
     {
-        $this->request = $request;
+        if ($requestStack instanceof Request) {
+            $this->request = $requestStack;
+        } elseif ($requestStack instanceof RequestStack) {
+            $this->request = $requestStack->getMasterRequest();
```
rybakit: It will not work for sub-requests, right? Don't know if it's important though.

jvandesande: It has never worked for sub requests. So it would be backwards incompatible if it would work now.

XWB: Other bundles use `getCurrentRequest()`.

stloyd: `getCurrentRequest()` can return `null`.

rybakit: @stloyd  The `Request` instance can be null as well: http://symfony.com/doc/2.3/cookbook/service_container/scopes.html#using-a-synchronized-service

PR #1032, opened 2016-03-01
Made default sort options from pagination config to work as expected

dariuszz123:
Fallback to defaultSortFieldName and defaultSortDirection pagination options instead of ignoring then.


XWB:
Thanks.

robhunt3r:
We are getting errors like `Undefined index defaultSortFieldName` even when we do not sort our results. We are using KnpPaginatorBundle, but it was working few days ago, we just updated FosElastica so problem may be here.

dariuszz123:
Lets revert this PR for now. I will inspect compatibility with KnpPaginatorBundle issues and create new PR.

XWB:
@dariuszz123 Done https://github.com/FriendsOfSymfony/FOSElasticaBundle/commit/0f13ecc916f8f7dde92f371acfdfc6d7e3925933

PR #873, opened 2015-04-28
Move provider options always used to the base AbstractProvider

merk:

PR #1038, opened 2016-03-02
Move provider options always used to the base AbstractProvider

XWB:
Replaces https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/873

### DIFF
--- a/Doctrine/AbstractProvider.php
+++ b/Doctrine/AbstractProvider.php
@@ -121,23 +121,6 @@ abstract class AbstractProvider extends BaseAbstractProvider
     }
 
     /**
-     * {@inheritDoc}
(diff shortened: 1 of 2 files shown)


merk:
I'm not sure I like this, but I'm sure I had a good reason originally. Seems to me like options specific to an implementation shouldnt have anything to do with the base class, and that any implementor should have specific options defined.

IIRC, it was because the Propel Provider was getting passed options it didnt know how to handle and it was going to be too difficult to fix that root cause.

XWB:
Closing this in favor of https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1043

PR #1039, opened 2016-03-02
Fixed documentation references

rvanlaarhoven:
Fixed the Dynamic Templates reference, which resulted in a 404.
Also updated the Date Format reference to the new domain.


XWB:
Thanks.

PR #1043, opened 2016-03-04
Fix Propel Config

sebastianhoitz:
We are using this at our company and we had to introduce these changes to make it work with Propel.


XWB:
We should have this fixed in 3.1, can you re-open this PR against the 3.1.x branch?

XWB:
Thanks.

PR #436, opened 2013-12-17
More code coverage

craigmarvelley:
Adds tests for

* TransformedFinder
* ModelToElasticaAutoTransformer
* ElasticaDataCollector
* RepositoryManager
* AbstractElasticaToModelTransformer

PR #1051, opened 2016-03-11
Improve code coverage

XWB:
Rebased from https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/436

### DIFF
--- a/Tests/Transformer/ModelToElasticaAutoTransformerTest.php
+++ b/Tests/Transformer/ModelToElasticaAutoTransformerTest.php
@@ -489,7 +489,7 @@ class ModelToElasticaAutoTransformerTest extends \PHPUnit_Framework_TestCase
         $document    = $transformer->transform(new POPO(), array(
             'upper' => array(
                 'type' => 'nested',
-                'properties' => array('name' => '~')
+                'properties' => array('name' => null)
             )
         ));
 
--- a/Tests/Transformer/ModelToElasticaAutoTransformerTest.php
+++ b/Tests/Transformer/ModelToElasticaAutoTransformerTest.php
@@ -503,7 +503,10 @@ class ModelToElasticaAutoTransformerTest extends \PHPUnit_Framework_TestCase
         $document    = $transformer->transform(new POPO(), array(
             'nullValue' => array(
                 'type' => 'nested',
-                'properties' => array()
+                'properties' => array(
+                    'foo' => array(),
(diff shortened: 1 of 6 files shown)


Tests/Transformer/ModelToElasticaAutoTransformerTest.php
```diff
@@ -473,6 +483,54 @@ public function testThatNestedObjectsDontNeedAnIdentifierField()
         ), $data['subWithoutIdentifier']);
     }
 
+    public function testNestedTransformHandlesSingleObjects()
+    {
+        $transformer = $this->getTransformer();
+        $document    = $transformer->transform(new POPO(), array(
+            'upper' => array(
+                'type' => 'nested',
+                'properties' => array('name' => null)
```
XWB: For some reason the original code `'properties' => array('name' => '~')` doesn't work on PHP 5.3, perhaps @stof has any idea?

stof: @XWB ``name: ~`` in YAML does not mean a ``'~'`` string after parsing. An unquoted ``~`` value is just an alternate syntax for ``null``. Is it your misunderstanding ?

XWB: @stof Have a look at the previous build:

https://travis-ci.org/FriendsOfSymfony/FOSElasticaBundle/builds/116368082

All tests except PHP 5.3 passed. It says:

> FOS\ElasticaBundle\Tests\Transformer\ModelToElasticaAutoTransformer\ModelToElasticaAutoTransformerTest::testNestedTransformHandlesSingleObjects
Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException: Neither the property "~" nor one of the methods "get~()", "~()", "is~()", "has~()", "__get()" exist and have public access in class "stdClass".

This was the original code:

```php
    public function testNestedTransformHandlesSingleObjects()
    {
        $transformer = $this->getTransformer();
        $document    = $transformer->transform(new POPO(), array(
            'upper' => array(
                'type' => 'nested',
                'properties' => array('name' => '~')
            )
        ));
        $data = $document->getData();
        $this->assertEquals('a random name', $data['upper']['name']);
    }
```

I'm trying to understand why it failed on PHP 5.3, and passed on PHP 5.4/5.5/5.6/7.

PR #1052, opened 2016-03-12
Search Elastic Search Results from a Repository

pmithrandir:
Hello,

It's been some time I'm wondering why getting hybrid or Doctrine Result would be done a way, and getting Elastic Search Result another way.
For performance reason, I decided to not use Doctrine Result neither the Hybrid one.

As I needed Pagination, I created all methods that will allow user to retrieve Raw Result, meaning an Array containing all Elastic search data.

Example of usage : 
In the repository

```
        $query =  new \Elastica\Query\MatchAll();
        return $this->findRawResult($query);
```

And with pagination

```
        $repositoryManager = $this->get('fos_elastica.manager.orm');
        $repository = $repositoryManager->getRepository('MyBundle:Entity');
        $pagerfanta = $repository->findRawPaginated($myQuery);
```

This change is only adding new functionalities, not affecting any other.

Some improvement could be done. For example, ElasticAdapter from PagerFanta Bundle, return Object you can use like that : 

```
foreach($results as $result){
    echo $result->getId();
}
```

The FantaPaginatorAdapter prevent that by forcing array to be returned in place of object. Could maybe be a good point to allow both.

### DIFF
--- a/Finder/TransformedFinder.php
+++ b/Finder/TransformedFinder.php
@@ -6,6 +6,7 @@ use Elastica\Document;
 use FOS\ElasticaBundle\Transformer\ElasticaToModelTransformerInterface;
 use FOS\ElasticaBundle\Paginator\TransformedPaginatorAdapter;
 use FOS\ElasticaBundle\Paginator\FantaPaginatorAdapter;
+use FOS\ElasticaBundle\Paginator\RawPaginatorAdapter;
 use Pagerfanta\Pagerfanta;
 use Elastica\SearchableInterface;
 use Elastica\Query;
@@ -46,7 +47,13 @@ class TransformedFinder implements PaginatedFinderInterface
 
         return $this->transformer->hybridTransform($results);
     }
+    
+    public function findRawResult($query, $limit = null, $options = array())
+    {
+        $results = $this->search($query, $limit, $options);
 
+        return $results;
+    }
     /**
      * Find documents similar to one with passed id.
      *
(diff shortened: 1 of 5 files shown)


Finder/TransformedFinder.php
```diff
@@ -97,6 +104,17 @@ public function findPaginated($query, $options = array())
...
+    
+    /**
+     * {@inheritdoc}
+     */
+    public function findRawPaginated($query, $options = array())
+    {
+        $queryObject = Query::create($query);
+        $paginatorAdapter = $this->createRawPaginatorAdapter($queryObject, $options);
+
+        return new Pagerfanta(new FantaPaginatorAdapter($paginatorAdapter));
```
XWB: I'm not entirely sure if this belongs in `TransformedFinder` as it doesn't really transforms something?

pmithrandir:
Hello,

The purpose of this code is to ease the integration of pure elastic search request, without having to change the entire code of your application.

For coherence, I put the new method in the same place as the one that does something(like transformation) to not diverge too much.

It seemed more logical and easier to understand.

Also, your note is a bit strange as the pagination is added. You don't count it as transformation of the result ?

Pierre

skonsoft:
👍

XWB:
@pmithrandir Can you update this PR and rebase with master?

XWB:
Closing as we want to do this differently in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/559

PR #1056, opened 2016-03-17
Add connectTimeout to options

ChristianRiesen:
To allow the use of connect timeout option. Default is from curl and thats 300 seconds, which makes near zero sense in any place where this might be used. So having the ability to set this correctly helps the client fail early if need be.


DependencyInjection/FOSElasticaExtension.php
```diff
@@ -110,6 +110,11 @@ private function loadClients(array $clients, ContainerBuilder $container)
             if (false !== $logger) {
                 $clientDef->addMethodCall('setLogger', array(new Reference($logger)));
             }
+
+            if (isset($clientConfig['connections'][0]['connectTimeout'])) {
+                $clientDef->addMethodCall('setConnectTimeout', [$clientConfig['connections'][0]['connectTimeout']]);
```
XWB: This won't work on PHP 5.3, please use `array()` instead.

XWB:
Can you extend the test as well?

ChristianRiesen:
Thanks for the feedback! Will do the changes during the day.

DependencyInjection/FOSElasticaExtension.php
```diff
@@ -110,6 +110,11 @@ private function loadClients(array $clients, ContainerBuilder $container)
             if (false !== $logger) {
                 $clientDef->addMethodCall('setLogger', array(new Reference($logger)));
             }
+
+            if (isset($clientConfig['connections'][0]['connectTimeout'])) {
```
merk: Why cant we support all the defined connections?

ChristianRiesen: This code was actually not needed. I dug a bit deeper and found that if a config value of connectTimeout is given, that does the trick already and it should set everything correctly inside the underlying elastica library.

ChristianRiesen:
@XWB Found an easier solution and added a test. Please have a look. Thanks!

merk:
Thanks!

PR #1059, opened 2016-03-29
Fix for UUID

netcreatoreu:
FOSElasticaBundle doesn't work correctly when we use UUID as primary key.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,13 +11,19 @@
         { "name": "Richard Miller", "email": "richard.miller@limethinking.co.uk" },
         { "name": "Jeremy Mikola", "email": "jmikola@gmail.com" }
     ],
+    "repositories": [
+        {
+            "type": "git",
+            "url": "https://github.com/netcreatoreu/Elastica"
+        }
+    ],
     "require": {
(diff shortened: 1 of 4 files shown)


merk:
An identifier should not be locked into being a string. 

Does your UUID class have a __toString() method? Otherwise, add a method 'getIdentifierString' and change the property path for the identifier in your type configuration.

damienalexandre:
Keep in mind that identifier are always strings in Elasticsearch.

composer.json
```diff
@@ -11,13 +11,19 @@
         { "name": "Richard Miller", "email": "richard.miller@limethinking.co.uk" },
         { "name": "Jeremy Mikola", "email": "jmikola@gmail.com" }
     ],
+    "repositories": [
+        {
+            "type": "git",
+            "url": "https://github.com/netcreatoreu/Elastica"
```
solivier: I don't think this should be merged

merk:
If ES identifiers are always strings then this isnt such a big deal. Since #1067 has test cases I'm happy to merge that one.

Issue #568, opened 2014-05-11
Cannot read property "id" from an array. Maybe you should write the property path as "[id]" instead?


ghost:
I'm getting this error message when ever I submit a search:

```
[2014-05-11 08:31:45] request.CRITICAL: Uncaught PHP Exception Symfony\Component\PropertyAccess\Exception\NoSuchPropertyException: "Cannot read property "id" from an array. Maybe you should write the property path as "[id]" instead?" at /home/kherrera/path/to/app/vendor/symfony/symfony/src/Symfony/Component/PropertyAccess/PropertyAccessor.php line 211 {"exception":"[object] (Symfony\\Component\\PropertyAccess\\Exception\\NoSuchPropertyException: Cannot read property \"id\" from an array. Maybe you should write the property path as \"[id]\" instead? at /home/kherrera/path/to/app/vendor/symfony/symfony/src/Symfony/Component/PropertyAccess/PropertyAccessor.php:211)"} []
```

This only occurs when I use `hydrate: false`, and it happens with both custom and default transformers.

From what I have been able to understand, some process in the FOSElasticaBundle is attempting to access an array (generated via array hydration) as an object. Does this sound like a bug, or is this an issue with how I configured the mapping? I am not using the JMS serializer.

PR #1062, opened 2016-04-07
array hydration fails to get the id property

MetalArend:
**Error: Cannot read property "id" from an array. Maybe you intended to write the property path as "[id]" instead.**

If I set the hydration option to false on the ElasticaToModelTransformer, the transform function complains that it cannot get the property id, because it does not detect "id" as an index inside the Symfony Component PropertyPathAccessor (or more correctly during the PropertyPath construction). I found issue #568, but the solution with the brackets throws the error:

**QueryException: SELECT o FROM XXX o WHERE o.[id] IN(:values)**

because of the line

`$qb->andWhere($qb->expr()->in(static::ENTITY_ALIAS . '.' . $this->options['identifier'], ':values'))`

inside the ElasticaToModelTransformer.

### DIFF
--- a/Doctrine/AbstractElasticaToModelTransformer.php
+++ b/Doctrine/AbstractElasticaToModelTransformer.php
@@ -96,7 +96,11 @@ abstract class AbstractElasticaToModelTransformer extends BaseTransformer
 
         // sort objects in the order of ids
         $idPos = array_flip($ids);
-        $identifier = $this->options['identifier'];
+        if ($this->options['hydrate']) {
+            $identifier = $this->options['identifier'];
+        } else {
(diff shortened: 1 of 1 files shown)


XWB:
Would it be possible to use the PropertyAccessor instead?

MetalArend:
@XWB The PropertyAccessor is already getting used by the getSortingClosure on the next line. For some reason the isIndex in PropertyAccessor is treating the objectOrArray in the wrong way.

XWB:
Should be fixed in https://github.com/FriendsOfSymfony/FOSElasticaBundle/commit/e9cb629caaee92920cb9810178ea12b3734b00a5

PR #1065, opened 2016-04-15
Update elastica lib

Gladhon:


solivier:
Hi @Gladhon,

it seems it's not compatible php3 and accordind to the .travis it has to be compatible.

Gladhon:
you're right. But php5.3 is no longer supported since 2014. Perhaps it can be dropped, symfony3 also dropped support for php5.3

solivier:
Agree the same for the support of php5.4

XWB:
Will be done in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1069

PR #1066, opened 2016-04-17
Update composer.json

jjborie:


solivier:
@jjborie same comment for the Pr https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1065 it does not works for php5.3 and the team want to keep the php5.3 version (cf. https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1002#issuecomment-191650487)

jjborie:
No problem, I will use my own branch with the patch.

PR #1067, opened 2016-04-19
Fixed ability to use an object as unique identifier

xavismeh:
When using a lib such as `ramsey/uuid-doctrine` to handle object IDs, the cast is not performed blocking transformation.

### DIFF
--- a/Tests/Transformer/ElasticaToModelTransformerCollectionTest.php
+++ b/Tests/Transformer/ElasticaToModelTransformerCollectionTest.php
@@ -100,6 +100,32 @@ class ElasticaToModelTransformerCollectionTest extends \PHPUnit_Framework_TestCa
         ), $results);
     }
 
+    public function testTransformOrderWithIdAsObject()
+    {
+        $this->collectionSetup();
+
+        $id1 = 'yo';
+        $id2 = 'lo';
+        $idObject1 = new IDObject($id1);
+        $idObject2 = new IDObject($id2);
+        $document1 = new Document($idObject1, array('data' => 'lots of data'), 'type1');
+        $document2 = new Document($idObject2, array('data' => 'not so much data'), 'type1');
+        $result1 = new POPO($idObject1, 'lots of data');
+        $result2 = new POPO2($idObject2, 'not so much data');
+
+        $this->transformers['type1']->expects($this->once())
+         ->method('transform')
+         ->with(array($document1, $document2))
+         ->will($this->returnValue(array($result1, $result2)));
+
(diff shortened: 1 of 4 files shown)


Tests/Transformer/ElasticaToModelTransformerCollectionTest.php
```diff
@@ -175,3 +201,21 @@ public function getId()
...
+    public function __construct($id)
+    {
+        $this->id = $id;
+    }
+
+    public function __toString()
+    {
+        return (string) $this->id;
+    }
+}
```
solivier: misssing blank line

solivier:
Hi @xavismeh I think you should check this comment https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1059#issuecomment-203098840

xavismeh:
Closing this PR as it duplicates https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1059

merk:
While its a duplicate, it does have test coverage.

Is this the best approach? Should we be casting ids to strings on every operation? Is it terrible to have a getStringIdentifier() method on the object instead of explicitly casting every return value of getId()?

xavismeh:
Here is the thought I had prior opening this PR and implementing that feature this way:
- elasticsearch identifiers will *always* be strings as mentioned in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1059#issuecomment-205018638
-  adding a custom method does not make sense to me as it breaks SoC: even if some think it's an antipattern, to me, the responsability for any given model class is to carry data through a piece of code. But should not have any clue on how to deal with it, since that logic should be moved somewhere else.

I have to admit however that in this case, casting to strings is not the **perfect** solution, it's "just" a good one: to get the perfect behavior, the propertyAccessor should be in charge of this. So maybe the proper solution would simply be to override the `identifier` configuration key at the application level instead to have  `id.toString` (this is the method used to cast Uuid objects) ?

XWB:
@xavismeh What do you think of https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1079 ?

xavismeh:
@XWB I don't get why there is a `cleanValue` method. ES requires a string, so in the property accessor class, I would simply have done as follow :
```php
// ...
public function getValue($objectOrArray, $propertyPath)
{
    return (string) parent::getValue($objectOrArray, $propertyPath);
}
```
Just because the rest of the code is useless IMHO !

Nicofuma:
Hello, quick question: why was it don on master only?

PR #1070, opened 2016-04-21
Add More Debugging Output

TheFox:
The original message doesn't actually help anybody. If this exception pops up provide the count of the Doctrine objects and the count of the Elastica results. Also print the IDs found by Elastica for further
investigation about the out-of-sync state.

### DIFF
--- a/Doctrine/AbstractElasticaToModelTransformer.php
+++ b/Doctrine/AbstractElasticaToModelTransformer.php
@@ -87,7 +87,7 @@ abstract class AbstractElasticaToModelTransformer extends BaseTransformer
         $objectsCnt = count($objects);
         $elasticaObjectsCnt = count($elasticaObjects);
         if (!$this->options['ignore_missing'] && $objectsCnt < $elasticaObjectsCnt) {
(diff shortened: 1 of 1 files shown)


Doctrine/AbstractElasticaToModelTransformer.php
```diff
@@ -84,8 +84,10 @@ public function transform(array $elasticaObjects)
         }
 
         $objects = $this->findByIdentifiers($ids, $this->options['hydrate']);
-        if (!$this->options['ignore_missing'] && count($objects) < count($elasticaObjects)) {
-            throw new \RuntimeException('Cannot find corresponding Doctrine objects for all Elastica results.');
+        $objectsCnt = count($objects);
+        $elasticaObjectsCnt = count($elasticaObjects);
+        if (!$this->options['ignore_missing'] && $objectsCnt < $elasticaObjectsCnt) {
+            throw new \RuntimeException('Cannot find corresponding Doctrine objects ('.$objectsCnt.') for all Elastica results ('.$elasticaObjectsCnt.'). IDs: '.join(', ', $ids));
```
solivier: you could use sprintf instead

TheFox: Yeah, acutally a clear case to use `sprintf`. 👍

XWB:
Thanks.

PR #1071, opened 2016-04-29
Removed immediate option in config

giovannialbero1992:
I've removed this option **immediate: true** from options of **listener**.

This option calls postFlush before any other doctrine event and avoids every update or insert of entity related to type of index.

i.e. If I declare that options such as

```
listener:
  immediate: true
 ```
on my index this will not be synchronized with entities persisted on my database

P.S. I've removed it also from documentation.

### DIFF
--- a/CHANGELOG-3.2.md
+++ b/CHANGELOG-3.2.md
@@ -16,3 +16,4 @@ CHANGELOG for 3.2.x
  * Resolved Propel configuration
  * Add Elastica compression option
  * Add support for `defaultSortFieldName` and `defaultSortDirection` pagination options
+ * Removed `immediate` option on type persistence configuration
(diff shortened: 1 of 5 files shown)


solivier:
it breaks a lot of tests

giovannialbero1992:
Thanks @solivier, I've removed the same commit from previous PR, so now are different PR. However I must find why the tests are red with SYMFONY 2.3.

XWB:
Symfony 2.3 error has been resolved https://github.com/FriendsOfSymfony/FOSElasticaBundle/commit/f8425d073d610ddb88d1795d10bddd639291acfe

giovannialbero1992:
Thanks @XWB and @solivier, I've rebased the PR.

XWB:
We need to add an entry to the changelog as well.

giovannialbero1992:
@XWB I've updated changelog

XWB:
Thanks.

PR #1073, opened 2016-05-05
add @Event annotation to let IDEs known event names and class instance

Haehnchen:
add @Event annotation for event names. see also https://github.com/symfony/symfony/commit/6857c36576b4885b88ef8b8784a7fc2551ba8962


merk:
Thanks!

PR #1077, opened 2016-05-18
Fix tests

solivier:
* Test were broken because of this PR introduced in symfony 2.3:
https://github.com/symfony/symfony/pull/18265

* We were passing an object instead of a string.


XWB:
Thanks.

PR #1079, opened 2016-05-23
Add ability to use an object as unique identifier

gmorel:
Rebound of #1067

Here is another proposal based on @xavismeh custom `PropertyAccessor` idea. Not sure if it is the best one though.

This implementation
- Cast to string only object having `__toString()` method
- Ignore objects not implementing `__toString()` method
- Ignore scalars (int, float, etc..)
- But performs a `is_object()` check on every `PropertyAccessor::getValue()` call.

----------------------

I tried implementing the `id.toString` concept without success: too messy .

----------------------

@merk I let you tell us in which direction you wish to go. 
Whatever you decide. I can give a hand implementing it.

@xavismeh Sorry for hijacking your Pull Request.

Keep up the good work !

### DIFF
--- a/PropertyAccessor/CastToStringPropertyAccessor.php
+++ b/PropertyAccessor/CastToStringPropertyAccessor.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace FOS\ElasticaBundle\PropertyAccessor;
+
+use Symfony\Component\PropertyAccess\PropertyAccessor;
+
+/**
+ * Allow to access a property value and directly call it's __toString() method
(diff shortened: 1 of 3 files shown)


merk:
I'm still :-1: on this - what I'm planning for 4.0 will not be compatible with objects for ids and it will be up to the/a serializer to handle it.

I agree its messy right now, because of the code soup. The only 'elegant' solution from a user perspective in my eyes is to implement 2 independent property paths, one for querying and one for serializing.

XWB:
Closing this, see above.

Issue #1083, opened 2016-05-25
Use default child node definition for index peristence provider, listener and finder


mickadoo:
Maybe this is just some oversight by me, but after going through the setup as described in the docs I noticed that I need to include set `provider`, `listener` and `finder` to the default `~` in the config.yml

I prefer to keep things minimal if possible, so I wondered why I couldn't reduce 

```
    ...
    indexes:
        app:
            types:
                story:
                    persistence:
                        driver: orm
                        model: MyBundle\Entity\Story
                        provider: ~
                        listener:
                            immediate: ~
                        finder: ~
```

to 

```
    ...
    indexes:
        app:
            types:
                story:
                    persistence:
                        model: MyBundle\Entity\Story
```

It's nice that it's so configurable but I don't see the point in required the configuration option to be set just to provide the default value.

XWB:
https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1084

PR #1084, opened 2016-05-25
[#1083] Make orm, provider, listener and finder optional. 

mickadoo:
As mentioned [in the issue](https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1083) I felt that these should be optional. 

2 of the tests are failing but these were failing before I made any changes.


XWB:
Thanks.

PR #1093, opened 2016-06-10
Ability to close index instead of delete

oleg-andreyev:
In case if you don't want to delete old index (backup for emergency case) you should close it.
Adding `--no-delete` option to `PopulateCommand`

### DIFF
--- a/Command/PopulateCommand.php
+++ b/Command/PopulateCommand.php
@@ -54,6 +54,7 @@ class PopulateCommand extends ContainerAwareCommand
             ->addOption('index', null, InputOption::VALUE_OPTIONAL, 'The index to repopulate')
             ->addOption('type', null, InputOption::VALUE_OPTIONAL, 'The type to repopulate')
             ->addOption('no-reset', null, InputOption::VALUE_NONE, 'Do not reset index before populating')
+            ->addOption('no-delete', null, InputOption::VALUE_NONE, 'Do not delete index after populate')
             ->addOption('offset', null, InputOption::VALUE_REQUIRED, 'Start indexing at offset', 0)
             ->addOption('sleep', null, InputOption::VALUE_REQUIRED, 'Sleep time between persisting iterations (microseconds)', 0)
             ->addOption('batch-size', null, InputOption::VALUE_REQUIRED, 'Index packet size (overrides provider config option)')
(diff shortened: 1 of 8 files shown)


Index/AliasProcessor.php
```diff
@@ -154,6 +162,30 @@ private function deleteIndex(Client $client, $indexName)
...
     /**
+     * Close an index
+     *
+     * @param Client $client
+     * @param string $indexName
+     */
+    private function closeIndex(Client $client, $indexName)
+    {
+        try {
+            $path = sprintf("%s/_close", $indexName);
```
oleg-andreyev: We should use: `\FOS\ElasticaBundle\Elastica\Index`.

EventListener/PopulateListener.php
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    /**
+     * @param IndexPopulateEvent $event
+     */
+    public function onPostIndexPopulate(IndexPopulateEvent $event)
+    {
+        $this->resetter->switchIndexAlias($event->getIndex(), $event->getOption('delete'));
+    }
+}
```
solivier: missing new line at the end of the file

oleg-andreyev: Don't know why github is trimming last line in diff, but if you'll checkout branch you'll see it's there.

oleg-andreyev:
@merk what do you think?

XWB:
Thanks.

Command/PopulateCommand.php
```diff
@@ -90,11 +91,16 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $index = $input->getOption('index');
         $type = $input->getOption('type');
         $reset = !$input->getOption('no-reset');
+        $delete = !$input->getOption('no-delete');
+
         $options = array(
+            'delete'        => $delete,
```
thanosp: Adding these as `$options` seems to break during the strict validation of options later on https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Provider/AbstractProvider.php#L65.
`reset` is already there but `delete` is missing.

oleg-andreyev: Ah. Seems that branch was quite old. I'll prepare fix for it ASAP

oleg-andreyev: @thanosp `default` options are defined in `configureOptions`
see https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Provider/AbstractProvider.php#L118

PR #1103, opened 2016-07-14
don't try to change number_of_shards on existing index

edast:
when resetting index type and applying settings - parameter number_of_shards should be skipped, as it cannot be changed for existing index - elasticsearch would throw exception:
`ElasticsearchIllegalArgumentException[can't change the number of shards for an index]`

was introduced in #911


XWB:
Thanks.

PR #1069, opened 2016-04-20
WIP - Compatibility with elasticsearch 2

giovannialbero1992:
I've modified code to work with elasticsearch 2.3.1 with elastica 3.1.1. 
This breaks BC in order to comply with ruflin/elastica >= 3.1.

It's working for all the uses in our application with no queries' modification (however raises some deprecation notices)

PR #1106, opened 2016-07-19
update Elastica new version

ReenExe:


XWB:
This won't work, see https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1069

PR #1107, opened 2016-07-22
fixing issue with _parent mapping and ES 2.3.3

oleg-andreyev:
This will fix issue with `Mapping definition for [_parent] has unsupported parameters:  [identifier : id] [property : null]`

### DIFF
--- a/DependencyInjection/FOSElasticaExtension.php
+++ b/DependencyInjection/FOSElasticaExtension.php
@@ -288,6 +288,11 @@ class FOSElasticaExtension extends Extension
                 $typeConfig['persistence'] = $type['persistence'];
             }
 
+            if (isset($type['_parent'])) {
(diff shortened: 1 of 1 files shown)


DependencyInjection/FOSElasticaExtension.php
```diff
@@ -2,13 +2,13 @@
 
 namespace FOS\ElasticaBundle\DependencyInjection;
 
-use Symfony\Component\HttpKernel\DependencyInjection\Extension;
-use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
```
solivier: Why did you remove this line ? It seems used in this class

oleg-andreyev: oops, it seems phpStorm had "Optimize imports" option enabled, reverted this part.

XWB:
Thanks.

oleg-andreyev:
Awesome! Thanks!

PR #1108, opened 2016-07-28
curl ssl option

xulien:
see https://github.com/ruflin/Elastica/issues/865


XWB:
Are we sure `CURLOPT_SSL_VERIFYPEER` is always available?

XWB:
Thanks.

PR #1114, opened 2016-08-09
Fixed ability to use object unique identifiers, continuing on 764ce12

morticue:
This is simply building on what was previously merged in PR #1067

This is doing the same, string-casting an object identifier, for a couple of cases which that PR did not cover.


XWB:
Thanks.

Issue #1100, opened 2016-06-24
parent-child mapping doesn't work, triggers exception when running 'populate'


rparsi:
Using:
- php 7.0
- symfony 3.0.x
- FOSElasticaBundle 'dev-master', which in turn uses Elastica 2.1
- elasticsearch 2.3

Using following mapping:
[https://gist.github.com/rparsi/7113704db69495274104316123e7d423](url)

When running `php bin/console fos:elastica:populate -vvv`
I get following:

`Resetting app
[2016-06-24 11:06:22] elastica.DEBUG: logging Request {"path":"app/","method":"DELETE","data":[],"query":[],"connection":{"config":{"headers":[]},"host":"localhost","port":9200,"logger":"fos_elastica.logger","compression":false,"retryOnConflict":0,"enabled":true}}
[2016-06-24 11:06:22] elastica.INFO: app/ (DELETE) 18.09 ms
[2016-06-24 11:06:22] elastica.DEBUG: logging Request {"path":"app/","method":"PUT","data":{"mappings":{"company":{"properties":{"id":{"type":"integer"},"name":{"type":"string","analyzer":"english"},"deleted":{"type":"boolean"}},"dynamic_date_formats":[],"_meta":{"model":"AppBundle\Entity\Company"}},"strain":{"properties":{"id":{"type":"integer"},"name":{"type":"string","analyzer":"english"},"deleted":{"type":"boolean"}},"dynamic_date_formats":[],"_meta":{"model":"AppBundle\Entity\Strain"}},"user":{"properties":{"id":{"type":"integer"},"username":{"type":"string","index":"not_analyzed"}},"dynamic_date_formats":[],"_meta":{"model":"AppBundle\Entity\User"}},"location":{"properties":{"id":{"type":"integer"},"address":{"type":"string"},"description":{"type":"string","analyzer":"english"},"geo":{"type":"geo_point"},"deleted":{"type":"boolean"}},"_parent":{"type":"company","property":null,"identifier":"id"},"dynamic_date_formats":[],"_meta":{"model":"AppBundle\Entity\Location"}}}},"query":[],"connection":{"config":{"headers":[]},"host":"localhost","port":9200,"logger":"fos_elastica.logger","compression":false,"retryOnConflict":0,"enabled":true}}

  [Symfony\Component\Debug\Exception\FatalThrowableError]
[...]

PR #1116, opened 2016-08-11
Restore property and identifier in _parent node mapping

alekitto:
Fixes a bug introduced in #1069 

`identifier` and `property` fields were removed instead of just being deleted from mapping array before sending it to ES


alekitto:
Probably fixes #1100 also

XWB:
Excellent, thanks!

PR #1118, opened 2016-08-12
Remove PHP 5.4 and Symfony 2.3 support

XWB:


composer.json
```diff
@@ -12,29 +12,31 @@
         { "name": "Jeremy Mikola", "email": "jmikola@gmail.com" }
     ],
     "require": {
-        "php": ">=5.4.0",
-        "symfony/framework-bundle": "~2.3|~3.0",
-        "symfony/console": "~2.1|~3.0",
-        "symfony/form": "~2.1|~3.0",
-        "symfony/property-access": "~2.3|~3.0",
+        "php": ">=5.5.0",
+        "symfony/framework-bundle": "~2.7|~3.0",
```
oleg-andreyev: why you've skipped 2.4, 2.5, 2.6, if PR was about 2.3?

mvrhov: AFAIK no. at least the whole Exception handling stuff doesn't work.

oleg-andreyev: Well 2.6.13 is 100% compatible (we've just switched to PHP7 :smile:)
Why I'm worried, it's because we're planning ElasticSearch upgrade to 2.x, but we are still on 2.6.13 (yes, it's not supported already)

mvrhov: IMO 2.6 to 2.7 should work better than 2.1 to 2.5 I was working on.

PR #1119, opened 2016-08-12
Repository Manager refactoring

alekitto:
i've ported the repositoryManager refactoring from our fork as requested in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1069#issuecomment-238535208

As stated in the original PR:
Currently there is no way to create more than one type associated with a model, each with its own repository class.
This PR adds this possibility, refactoring the RepositoryManager class

Now a repository can be retrieved through the `fos_elastica.repository_manager` service calling getRepository method with `index/type` string as first argument.
Old doctrine repository manager class has been deprecated, but are still working as expected.

BC breaks are present ONLY if the doctrine repository manager has been extended, while usage remains the same

### DIFF
--- a/DependencyInjection/FOSElasticaExtension.php
+++ b/DependencyInjection/FOSElasticaExtension.php
@@ -617,13 +617,18 @@ class FOSElasticaExtension extends Extension
             $container->setDefinition($finderId, $finderDef);
         }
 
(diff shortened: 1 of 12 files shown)


bobvandevijver:
I really wonder why this is merged without any documentation updates or migration guides.

PR #1128, opened 2016-08-18
[BUG] using --no-reset without --type

oleg-andreyev:
Accidentally discovered annoying bug.  
We have "large" dev-index (~12Gb ~6 millions docs) and when we need to update N documents we use `--no-reset` with `--type`, but today when running update `--type` wasn't added and `PopulateCommand` delete current index. :(

Fix: do not `postPopulate`, if `--no-reset` is passed.

### DIFF
--- a/Command/PopulateCommand.php
+++ b/Command/PopulateCommand.php
@@ -151,7 +151,7 @@ class PopulateCommand extends ContainerAwareCommand
 
         $this->dispatcher->dispatch(IndexPopulateEvent::POST_INDEX_POPULATE, $event);
 
-        $this->refreshIndex($output, $index);
+        $this->refreshIndex($output, $index, !$reset);
     }
 
     /**


Command/PopulateCommand.php
```diff
@@ -151,7 +151,7 @@ private function populateIndex(OutputInterface $output, $index, $reset, $options
 
         $this->dispatcher->dispatch(IndexPopulateEvent::POST_INDEX_POPULATE, $event);
```
oleg-andreyev: Don't we want to move
```
if ($postPopulate) {
             $this->resetter->postPopulate($index);
         }
```

into listener?

Listener was already created in #744, but than removed in 4c87d24fc12d3e08580bd6aecbfdd3b289e316f2

oleg-andreyev:
// @XWB

XWB:
@oleg-andreyev Moving stuff to a listener can be done in another PR.

XWB:
Thanks.

PR #1134, opened 2016-08-24
Fix property access error using hydrate false option

### LINKED ISSUE #1133: Transformer fail with false hydrate option
Cannot read property "id" from an array. Maybe you intended to write the property path as "[id]" instead.
vendor/friendsofsymfony/elastica-bundle/Doctrine/AbstractElasticaToModelTransformer.php at line 114

istvankis:
Fixes #1133


XWB:
Thanks.

PR #1138, opened 2016-09-04
[BUG] UndefinedOptionsException: The options "reset" do not exist.

oleg-andreyev:
When using `\FOS\ElasticaBundle\Provider\AbstractProvider` (Doctrine ORM)

```
[Symfony\Component\OptionsResolver\Exception\UndefinedOptionsException]
  The options "reset" do not exist. Defined options are: "batch_size", "clear_object_manager", "debug_logging", "igno
  re_errors", "indexName", "offset", "query_builder_method", "skip_indexable_check", "sleep", "typeName".

```

### DIFF
--- a/Provider/AbstractProvider.php
+++ b/Provider/AbstractProvider.php
@@ -105,6 +105,7 @@ abstract class AbstractProvider implements ProviderInterface
     protected function configureOptions()
     {
         $this->resolver->setDefaults(array(
+            'reset' => true,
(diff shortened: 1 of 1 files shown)


XWB:
Wouldn't it be better to make it default false?

oleg-andreyev:
By default command resets index (creating new one)

XWB:
Thanks.

Issue #1026, opened 2016-02-15
[Question] Is there a possibility to deactivate listener.immediate temporarily?


OskarStark:
Hi Guys

i would like to deactivate the listener during a long import task, and run `app/console fos:elastica:populate` afterwards instead.

is there a current solution for this?

thank you :+1:

XWB:
The `immediate` options has now been removed.

PR #1145, opened 2016-09-21
Global flag to disable object indexing (for long import tasks)

a-ast:
| Q | A |
| --- | --- |
| Bug fix? | no |
| New feature? | yes |
| BC breaks? | no |
| Deprecations? | no |
| Tests pass? | yes |
| Fixed tickets | #1026 |

We have a legacy script that creates and updates entities using Doctrine but also inserts new records with the QueryBuilder. At the end we get partially indexed data and have to run fos:elastica:populate to index it properly.
Moreover, old scripts works now significantly slower because of the Listener in FOSElasticaBundle.
In our case it would be much better to disable object indexing for console scripts and run reindexing manually when we ready for it.
My proposal is to disable Listener by extending the Indexer:

``` php
// In a console script before operations with EntityManager
$indexable = $this->getContainer()->get('fos_elastica.indexable');
$indexable->setIndexingEnabled(false);

// Create entities, persist remove, flush...
```

### DIFF
--- a/Doctrine/Listener.php
+++ b/Doctrine/Listener.php
@@ -143,6 +143,11 @@ class Listener
      */
     private function persistScheduled()
     {
+        // Don't persist if indexing is disabled
+        if (!$this->indexable->isIndexingEnabled()) {
+            return;
+        }
+
         if (count($this->scheduledForInsertion)) {
             $this->objectPersister->insertMany($this->scheduledForInsertion);
             $this->scheduledForInsertion = array();
--- a/Provider/Indexable.php
+++ b/Provider/Indexable.php
@@ -54,6 +54,11 @@ class Indexable implements IndexableInterface
     private $propertyAccessor;
 
     /**
+     * @var bool
+     */
+    private $indexingEnabled = true;
+
+    /**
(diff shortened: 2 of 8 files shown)


OskarStark:
any news on this PR?

XWB:
I'm afraid I do not fully understand the purpose of this PR?

a-ast:
@XWB The purpose has been explained in #1026 and in the description of this PR.
If there is a better way to achieve the same goal, it would be great to know how to do it. Then I'd write a cookbook page instead.
Your comment in the issue #1026
> The immediate options has now been removed.
doesn't make me clear how to easily deactivate listener. Could you explain please?

XWB:
@a-ast Ok, I see.

Disabling an event listener should be done in `FOSElasticaExtension`. Changing an interface is not the right approach to achieve this goal.

XWB:
Have a look at https://github.com/FriendsOfSymfony/FOSUserBundle/blob/master/DependencyInjection/FOSUserExtension.php on how to disable event listeners. I'm closing this PR in the meantime.

Issue #1142, opened 2016-09-13
fos_elastica.post_transform not working with serializer


eloirobe:
Hello

I Have an issue when I use the serializer option. When the serializer is disabled the fos_elastica.post_transform event is called but when I enable the serializer option, the post_transform is not called.

These are my files:

```
fos_elastica:
    clients:
        default: { host: %elastic_host%, port: %elastic_port% }
->  serializer:
->    callback_class: FOS\ElasticaBundle\Serializer\Callback
->    serializer: jms_serializer

    indexes:
        lyra:
            use_alias: true
            settings:
                index:
                    analysis:
                        analyzer:
                            my_analyzer:
                                type: custom
                                tokenizer: keyword
                                filter: [lowercase, asciifolding]
            types:
                expediente:
                    indexable_callback: 'isEnabled'
                    mappings:
                        id: ~
                        barcode: ~
                    persistence:
                        driver: orm
                        model: Domain\Entity
                        provider: ~
                        listener: ~
                        finder: ~
                    serializer:
                        groups: ['elastica']
```

```
class CustomPropertyListener implements EventSubscriberInterface
{

    public function addCustomProperty(TransformEvent $event)
    {
            $document = $event->getDocument();
            $document->set('newkey', 'value');
    }

    public static function getSubscribedEvents()
    {
        return array(
            TransformEvent::POST_TRANSFORM => 'addCustomProperty',
        );
    }
}

```

```
[vagrant@dev]$ composer -i show | grep -e "elastic\|serializer\|symfony"
You are using the deprecated option "installed". Only installed packages are shown by default now. The --all option can be used to show all packages.
behat/symfony2-extension                2.1.1              Symfony2 framework extension for Behat
friendsofsymfony/elastica-bundle        3.2.0              Elasticsearch PHP integration for your Symfony2 project using Elastica
friendsofsymfony/rest-bundle            1.7.7              This Bundle provides various tools to rapidly develop RESTful API's with Symfony
friendsofsymfony/user-bundle            dev-master 4cb7a29 Symfony FOSUserBundle
jms/serializer                          1.3.1              Library for (de-)serializing data of any complexity; supports XML, JSON, and YAML.
jms/serializer-bundle                   1.1.0              Allows you to easily serialize, and deserialize data of any complexity
ruflin/elastica                         2.3.1              Elasticsearch Client
symfony/assetic-bundle                  v2.8.0             Integrates Assetic into Symfony2
[...]

PR #1146, opened 2016-09-21
Dispatch event POST_TRANSFORM when serializer option is used

giovannialbero1992:
#1142

Now when serialize option is used the listener on `POST_TRANSFORM` event will be called.

I don't know if this PR is welcome because the data is serialized and is not possible to do something like this

``` php
$document = $event->getDocument();
$document->set('newkey', 'value');
```

but you must parse data and set it again into document, like in the test introduced with this PR.

``` php
$document = $event->getDocument();
$data = $document->getData();
if (is_string($data)) {
  $unserializeData = json_decode($data, true);
  $unserializeData['field1'] = 'post_persister';
  $document->setData(json_encode($unserializeData));
}
```

@eloirobe @AmOrFeU86
I recommend to try this fork on your private project to receive some more feedback here.
Thanks a lot


XWB:
While the idea is good, the implementation is way too cumbersome.

PR #1149, opened 2016-09-29
Execution time in profiler is not accurate 

oleg-andreyev:
I've noticed that execution time in profiler is wrong:

- Removed `$start` and `$end` from overridden `Client` in favor of `getQueryTime`
- Code Style fixes and improved type hinting in `logQuery`
- Cleaned `ElasticaLogger` by extending `AbstractLogger` which already implemented all methods required by interface

Before:
image
image

After:
image
image

### DIFF
--- a/Elastica/Client.php
+++ b/Elastica/Client.php
@@ -43,14 +43,13 @@ class Client extends BaseClient
             $this->stopwatch->start('es_request', 'fos_elastica');
         }
 
-        $start = microtime(true);
         $response = parent::request($path, $method, $data, $query);
         $responseData = $response->getData();
 
         if (isset($responseData['took']) && isset($responseData['hits'])) {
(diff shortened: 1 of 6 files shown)


Elastica/Client.php
```diff
@@ -43,14 +43,13 @@ public function request($path, $method = Request::GET, $data = array(), array $q
             $this->stopwatch->start('es_request', 'fos_elastica');
         }
 
-        $start = microtime(true);
         $response = parent::request($path, $method, $data, $query);
         $responseData = $response->getData();
```
oleg-andreyev: Before `getData` does a magic, executing time was inaccurate

oleg-andreyev:
ping @XWB

XWB:
Thanks.

PR #1153, opened 2016-10-05
Update Provider.php

samanmohamadi:
Declaration of FOS\ElasticaBundle\Doctrine\PHPCR\Provider::createQueryBuilder($method) must be compatible with FOS\ElasticaBundle\Doctrine\AbstractProvider::createQueryBuilder($method, array $arguments = Array)


XWB:
Thanks.

Issue #1156, opened 2016-10-15
Cast id to string not working


spiroski:
Hello I'm trying to map a MongoDB Doctrine entity that has an object ( Wrapper of UUID) as Id.

Here's my mapping:

```
types:
                employee:
                    mappings:
                        first_name:
                            property_path: name.first
                        last_name:
                            property_path: name.last
                        middle_name:
                            property_path: name.middle
                        email: ~
                        skills:
                            type: 'nested'
                            properties:
                                id:
                                    property_path: id.toString
                                names:
                                    type: 'nested'
                                    properties:
                                        locale: ~
                                        name: ~
                    persistence:
                        driver: mongodb
                        model: TwoBit\Employees\Model\Employee
                        provider: ~
                        listener: ~
                        finder: ~
                        identifier: id
```

Here's my entity:

```
class Employee implements Person, HasSkills, WorksInDepartments, Subordinate , Superior
{
    /**
     * @var string
     */
    protected $id;

    /**
     * @return PersonID
     */
    public function getId()
    {
        return PersonID::fromString($this->id);
    }

}
```

PersonID overrides the __toString() method and returns a valid string representation of UUID.

The private property id is a string but I cast it to PersonID when using the getter. 

From searching the issues I understand that this has been fixed before, but it doesn't seem to work with "dev-master" version.

PR #1157, opened 2016-10-16
Cast identifier to string in case it is an object

spiroski:
If the entity uses object that can be casted to string, the bundle should do that.

Re: @ #1156


spiroski:
Can I have some kind of response regarding this issue? It has a big impact on my use-cases.

XWB:
Thanks.

XWB:
Reverting as this breaks as lot of tests.

spiroski:
I don't think reverting this back is the correct decision.

Do you agree that it's expected behavior for this bundle to handle object identifiers that are castable to string?
Will you merge this feature back in if I manage to fix the test problems?

PR #1161, opened 2016-10-27
Fix MongoDB config

laurentzoz:
I get this exception with symfony2 using doctrine and mongodb:

`Fatal Error: Argument 2 passed to FOS\ElasticaBundle\Doctrine\RepositoryManager::__construct() must implement interface FOS\ElasticaBundle\Manager\RepositoryManagerInterface, instance of Doctrine\Common\Annotations\CachedReader given`

I don't know if this issue needs more code but it seems to work.


XWB:
Thanks.

PR #1162, opened 2016-10-28
Support nested_path and nested_filter sort attributes when using knp pager

### LINKED ISSUE #1159: Nested Sort not Working on Paginate
I have the following mapping:

```
property:
   mappings:  
      data:
         type: nested
            properties:
               rooms: { type: integer }
               baths: { type: integer }
               size: { type: float }

```

I need to sort by size and i put the "nested_field" necessary like the elasticsearch guide says [Sorting within nested objects.](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-sort.html#nested-sorting)

When i add the sort array:
`"data.size" => array(
    "order" => "asc",
    "nested_path" => "data"
  )`

The paginator returns the query without nested_path element.

alekitto:
Should resolve #1159

Added support for `sortNestedPath` and `sortNestedFilter` options to knp paginator `paginate` method.
Allows to specify `nested_path` and `nested_filter` sort fields when paginating, passing a string or a callable accepting the current sort field.

### DIFF
--- a/Resources/doc/usage.md
+++ b/Resources/doc/usage.md
@@ -31,6 +31,20 @@ $countOfResults = $userPaginator->getNbResults();
 $paginator = $this->get('knp_paginator');
 $results = $finder->createPaginatorAdapter('bob');
 $pagination = $paginator->paginate($results, $page, 10);
+
+// You can specify additional options as the fourth parameter of Knp Paginator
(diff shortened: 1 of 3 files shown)


XWB:
Thanks.

PR #1163, opened 2016-10-31
Update PopulateCommand.php

fedotovaleksandr:
DialogHelper is depricated in Symfony ^3.*
In result crash on populate command with --offset , but without --no-reset options.
Use QuestionHelper


XWB:
Thanks.

PR #1164, opened 2016-11-07
Fix resetIndexType on ES 1.x

Tobion:
In version 1 of ES, the settings are different. Also the number_of_replicas cannot be changed on closed indexed as well:

> Can't update [index.number_of_replicas] on closed indices [] - can leave index in an unopenable state


XWB:
Thanks.

Issue #914, opened 2015-08-02
PopulateCommand with --offset option always displays progess at index 0 instead of given offset


rushi:
When you run the populate command:

```
php app/console fos:elastica:populate --type=foo--index=bar --no-reset --offset=200000
```

The progress bar output always begins at `0` instead of the given offset of `200000`, which causes incorrect calculation of the percentage value. 

**Expected Output:** 
`Populating search/transaction 50.0% (200100/400000), 100 objects/s (RAM: current=17Mo peak=19Mo)`
**Actual Output:** 
`Populating search/transaction 0.0% (100/400000), 100 objects/s (RAM: current=17Mo peak=19Mo)`

This occurs on the latest release of the bundle v3.1.4 using **Symfony 2.3**. 

Note: I've tested that this fails with the legacy progress (`\FOS\ElasticaBundle\Command\ProgressClosureBuilder::buildLegacy`). I can't test it with the new Symfony ProgressBar component

jeroenjoosen:
👍  -  Same issue here, that's why we thought that it didn't even work...

PR #1166, opened 2016-11-21
Fixed PopulateCommand when --offset has been specified

IonBazan:
This fixes the  #914 issue. ProgressBar now starts at the position of the specified offset. I was also considering if it should start at 0 with max value decreased by offset value.


XWB:
Thanks.

PR #1167, opened 2016-11-21
optimized travis.yml for faster builds

IonBazan:
This should significally decrease time spent on dependency installation by composer


alekitto:
No code coverage is generated after this change. PHPUnit warns that `No code coverage driver is available`.
Probably the `xdebug.ini` readded with `config-add` is not containing `extension=xdebug.so`

IonBazan:
You are right. I assumed that `phpenv config-rm` disables the config, but it actually removes it.

alekitto:
@IonBazan You can try this solution: https://github.com/Shippable/support/issues/2140#issuecomment-165718472

XWB:
Will close this, feel free to open a new PR.

PR #1168, opened 2016-11-22
Drop support for symfony < 2.7 - 2nd part

alekitto:
Use `ContainerAwareInterface` and `ContainerAwareTrait` in `Indexable` and `ProviderRegistry` classes, since we dropped support for sf < 2.7 and php < 5.5
Also remove `RequestCompilerPass`, since `request_stack` service should always be available


XWB:
:+1:

PR #1171, opened 2016-11-25
[WIP] Elasticsearch 5.x support

alekitto:
Add ES 5.x support via Elastica 5.0
Currently testing with `ruflin/elastica` beta version.

See also #1170

PR #1193, opened 2017-01-26
Upgrade elastica to ^5.1 and update build to match ES version

mweibel:
partially based on #1171


composer.json
```diff
@@ -12,13 +12,13 @@
         { "name": "Jeremy Mikola", "email": "jmikola@gmail.com" }
     ],
     "require": {
-        "php": ">=5.5.0",
-        "symfony/framework-bundle": "~2.7|~3.0",
-        "symfony/console": "~2.7|~3.0",
-        "symfony/form": "~2.7|~3.0",
-        "symfony/property-access": "~2.7|~3.0",
+        "php": ">=5.6.0",
+        "symfony/framework-bundle": "~2.7|~3.1",
```
thePanz: Can you use `||` to separate different version numbers?
https://getcomposer.org/doc/articles/versions.md#range

mweibel: fixed, thanks!

mweibel:
@XWB we've seen there's a CHANGELOG for 4.0 but a version 4.0 has never been released so far. Do you plan on releasing that or can we take 4.0 over with support for elastica lib 5.x?

XWB:
@mweibel The problem is that the current master branch (4.0-dev) is not fully compatible with ES 2. So if we tag 4.0, it would be a bugged release.

bitgandtter:
The big issue here is the impossibility of use this bundle with elastic 5.x, we dont need to address the improvements of this release and that even on the official site they encourage us to upgrade to this new version. No matter if it is a new project or an existing one we are all migrating to 5.x and now we have to deal with using or other bundles/libraries or fork this just to change restrictions on the composer.json because it works just fine just like it is. 

I understand that the project wants to keep things organized but you dont have the men power to release a fully compatible branch for 2.x and 5.x in the near future, so it will be really nice that the project decide to even create a new branch and tag it only for 5.x even if there is no much support to it but we can keep our integrations and work. As i mention i fork this on my own i just change the version restriction and its working just fine.

Please consider since the community its really needing this. Thanks

mweibel:
@XWB this PR should be ready now. Please review :)

.travis.yml
```diff
@@ -4,39 +4,31 @@ cache:
...
-      - elasticsearch-2.x
-    packages:
-      elasticsearch
+dist: trusty
+sudo: required
 
 matrix:
   include:
-    - php: 5.5
-      env: COMPOSER_FLAGS="--prefer-lowest --prefer-stable"
```
dbu: you need to keep a build with the lowest versions

CHANGELOG-4.0.md
```diff
@@ -6,14 +6,16 @@ in 4.0 versions.
 
 * 4.0.0 (xxxx-xx-xx)
 
- * Add `ruflin/Elastica` 3.x support.
```
dbu: note, the ruflin/elastica 3.x support is already tagged in this bundle.

CHANGELOG-4.0.md
```diff
@@ -6,14 +6,16 @@ in 4.0 versions.
...
+ * Add `ruflin/Elastica` 5.1 support.
```
dbu: i would go with the composer name ruflin/elastica (lowercase) https://github.com/ruflin/Elastica/blob/07a770164b37716748aca1f19340cce48233229b/composer.json#L2

composer.json
```diff
@@ -12,31 +12,31 @@
...
-        "php": ">=5.5.0",
-        "symfony/framework-bundle": "~2.7|~3.0",
-        "symfony/console": "~2.7|~3.0",
-        "symfony/form": "~2.7|~3.0",
-        "symfony/property-access": "~2.7|~3.0",
+        "php": ">=5.6.0",
+        "symfony/framework-bundle": "~2.7 || ~3.1",
+        "symfony/console": "~2.7 || ~3.1",
+        "symfony/form": "~2.7 || ~3.1",
+        "symfony/property-access": "~2.7 || ~3.1",
```
dbu: lets change these from ~ to ^

CHANGELOG-4.0.md
mweibel: fixed

Tobion:
👍 for merging

CHANGELOG-4.0.md
Tobion: @dbu I don't think it is. But there doesn't seem much damand. So I don't think it matters now.

dbu: ups, you are right, i got confused between the 3.2.x branch and master. if people relied on dev-master or 4.0 @ dev then making that suddenly be for ES 5, it could create confusion. maybe we could create a tag as es-3-best-effort or something like that, so that at least that state can be references if somebody needs it.

XWB:
I just created a 4.0.x branch as we need to maintain Elasticsearch 2 support. Please move the changelog entries in this PR to `CHANGELOG-5.0.md`.

mweibel:
@XWB ok. I rebased my branch on current master and adjusted CHANGELOG/UPGRADE files.

dbu:
@XWB so the 4.x branch would be about ES 3, not ES 2 right? ES 2 is in the 3.x branch, what was master until now is for ES 3 according to the changelogs and composer dep on ruflin/elastica.

i think that makes sense. it will be quite some effort to maintain 3 versions. do you plan to put the 3.x and 4.x branches into maintainance mode and new features only get added to 5?

mfeltscher:
@dbu What do you mean by "ES 3"? There is no Elasticsearch version 3.`ruflin/elastica` version 3 is about Elasticsearch version 2.

dbu:
> @dbu What do you mean by "ES 3"? There is no
> Elasticsearch version 3.|ruflin/elastica| version 3 is about
> Elasticsearch version 2.

ugh, sorry. i understood that ruflin/elastica replicates the ES version
it works for. so version 4.x supports ES 2 with the refactored
ruflin/elastica library version 3.

mfeltscher:
@dbu No worries :) At least starting with v5 `ruflin/elastica` replicates the Elasticsearch version again :)

XWB:
@dbu Yes we should keep Elasticsearch 2 (= Elastica 3) support in the 4.x branch.

XWB:
Thanks.

XWB:
@mweibel @dbu Do we also need to remove `_timestamp` and `_ttl` from https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/DependencyInjection/FOSElasticaExtension.php#L250 ?

Tobion:
@XWB yes indeed. Good catch. Will you make the changes?

bitgandtter:
Im getting the following exception after the merge
                                                                                                                                                              
```
  [Symfony\Component\OptionsResolver\Exception\UndefinedOptionsException]                                                                                     
  The option "delete" does not exist. Defined options are: "batch_size", "clear_object_manager", "debug_logging", "ignore_errors", "indexName", "offset", "q  
  uery_builder_method", "reset", "skip_indexable_check", "sleep", "typeName".                                                                                 
                                                                                                                                                              

fos:elastica:populate [--index [INDEX]] [--type [TYPE]] [--no-reset] [--no-delete] [--offset OFFSET] [--sleep SLEEP] [--batch-size BATCH-SIZE] [--ignore-errors] [--no-overwrite-format] [-h|--help] [-q|--quiet] [-v|vv|vvv|--verbose] [-V|--version] [--ansi] [--no-ansi] [-n|--no-interaction] [-e|--env ENV] [--no-debug] [--] <command>
```

Dont know if its related

XWB:
@Tobion Will do.
@bitgandtter This has been fixed in master.

XWB:
We still have 2 failing tests in master. Anyone?

PR #1197, opened 2017-02-10
Fix call of unknown method QuestionHelper::askConfirmation()

fabianerni:
QuestionHelper has no method askConfirmation(), use ask()
Problem is caused by migration from DialogHelper to QuestionHelper in this commit: https://github.com/FriendsOfSymfony/FOSElasticaBundle/commit/9824b56143f448979430f764508dddb4231e072e


XWB:
Thanks.

PR #1199, opened 2017-02-16
Remove never executed code from Resetter

mephir:
Settings variable is not defined before condition. ```!empty($settings)``` will always return ```false``` which cause this code be never executed.


XWB:
@mephir Can you rebase from master?

mephir:
@XWB done

XWB:
Thanks.

PR #1208, opened 2017-02-28
remove all *.class parameters

Tobion:
To follow symfony best practices, see https://github.com/symfony/symfony/pull/14070


XWB:
Thanks :+1:

PR #1214, opened 2017-03-06
Add support for suggesters

XWB:


trsteel88:
getSuggests() always seems to return an empty array for me. Is there something in the query you need to include so elasticsearch returns suggestions?

pittersnider:
UP

PR #1215, opened 2017-03-06
Clone $query  because size is set to 0 in ruflin package, causing 0 results

whoamicz:
$paginator = $finder->findPaginated($boolQuery);
$paginator->setMaxPerPage(25);
$paginator->setCurrentPage($page);

empty because in ruflin/elastica/lib/Elastica/Search.php is

$query = $this->getQuery();
$query->setSize(0);

### DIFF
--- a/Paginator/RawPaginatorAdapter.php
+++ b/Paginator/RawPaginatorAdapter.php
@@ -118,7 +118,8 @@ class RawPaginatorAdapter implements PaginatorAdapterInterface
     public function getTotalHits($genuineTotal = false)
     {
         if (!isset($this->totalHits)) {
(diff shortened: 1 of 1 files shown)


XWB:
I'd like to understand why Elastica is setting size to zero, ping @ruflin

ruflin:
I assume this is related to https://github.com/ruflin/Elastica/issues/1261 ? BTW which version of Elastica is this?

XWB:
@ruflin I looks to be the same issue indeed.

XWB:
Fixed in https://github.com/ruflin/Elastica/pull/1264#event-1008219656

PR #1229, opened 2017-03-31
Pagination fix

jorge07:
Problem:

elastic: 5.2

I have a custom implementation with this method:

```
    private function pagerAdapter(Search $searchQuery): Pagerfanta
    {
        $paginatorAdapter = new FantaPaginatorAdapter(
            new TransformedPaginatorAdapter($this->index, $searchQuery->getQuery(), $this->options, $this->hydrator)
        );

        return new Pagerfanta($paginatorAdapter);
    }
```

Going to page 1 work well, but the second fails with '$itemCountPerPage must be greater than zero'. 

That happened because is looking for the size in the query instead `totalHits`, which it's the one needed in this case. See https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Paginator/RawPaginatorAdapter.php#L82

I'll try to add some tests to explain it better.


z38:
Elastica recently fixed an issue where the parameter `size` was overwritten inadvertently (ruflin/elastica#1264). Does the issue still exist if you upgrade `ruflin/elastica` to latest commit (`composer require "ruflin/elastica:dev-master"`)?

jorge07:
Yes, it look similar but are different, I still having the issue.

XWB:
This should be fixed in Elastica.

jorge07:
@XWB what it's failing in this case it's the pagination logic not elastica I think. But if I'm wrong can you give me some idea about how to deal with that issue?

PR #1232, opened 2017-04-07
Post populate event fix

kbsali:
Only a little update of the sample postPopulate trick not working with ES > V2.0


XWB:
I believe this should be merged in the 4.x branch.

kbsali:
true

PR #1236, opened 2017-04-18
Fix undefined index when ignore_missing is active

enumag:
Sending this both here and to FazlandElasticaBundle. https://github.com/fazland/FazlandElasticaBundle/pull/18


Transformer/ElasticaToModelTransformerCollection.php
```diff
@@ -97,6 +97,9 @@ public function hybridTransform(array $elasticaObjects)
 
         $result = [];
         for ($i = 0, $j = count($elasticaObjects); $i < $j; ++$i) {
+            if (! isset($objects[$i])) {
```
XWB: Could you change it to `!isset`?

Issue #689, opened 2014-08-11
[RFC] New pagination


merk:
We've got a few outstanding PRs to add aggregations or other features available to Elastica into our Pagination interfaces, but to me this exposes that the interfaces of this bundle are not flexible and need to be improved.

I propose the following way I believe that pagination should work:

``` php

$search = new \Elastica\Search($this->container->get('fos_elastica.client'));
$search->addIndex($this->container->get('fos_elastica.index.index_name'));
$search->addType($this->container->get('fos_elastica.index.type_name1'));
$search->addType($this->container->get('fos_elastica.index.type_name2'));

$query = /** Build an instance of an \Elastica_Query object that you want to search against the Searchable */

// Using Pagerfanta:
$adapter = new ElasticaAdapter($search, $query);
$pagerfanta = new Pagerfanta($adapter);

// Using Knp Paginator
$paginator  = $this->get('knp_paginator');
$pagination = $paginator->paginate(
    new ElasticaAdapter($search, $query),
    $this->get('request')->query->get('page', 1)/*page number*/,
    10/*limit per page*/
);
```

It may be possible to adjust the api slightly to avoid the need to create a Search object and accept a SearchableInterface object that we would convert internally into a Search object.

I also face problems with this implementation - the return values that we can return. For 3.2 or 4.0 I plan on removing our Hybrid results completely, and instead only return the Elastica ResultSet containing only Elastica Result objects. I hope to have a method, getTransformed() on the Result to return the transformed object. This cant be done for 3.1.

The only option I can think of that lets me keep some level of forward compatibility without forcing you all to make changes is:

``` php
// In php
foreach ($pagerfanta as $result) {
  $transformed = $resultTransformer->transform($pagerfanta, $result);

 // Do stuff with $transformed
}

// In twig
{% for result in pagerfanta %}
  {% set transformed = result | fos_elastica_transform(pagerfanta) %}

  {# Do stuff with transformed #}
{% endfor %}
```

I would appreciate feedback from those that actually use pagination, since in my applications I dont paginate search results.

nurikabe:
+1

hugohenrique:
I tested its proposal with a structure:
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-metrics-top-hits-aggregation.html#_example

And it did not work, you suggest something?

PR #1240, opened 2017-05-05
Split provider responsibility.

makasim:
The PR introduce new provider API. Everything related to new API is prefixed with `pager`. In 1.4.x the API is available but disabled by default (because of BC promise). To enable it set `pager_provider` option and `fos_elastica.use_v5_api` tot true. Like this:

```yaml
# app/config/config.yml

parameters:
    fos_elastica.use_v5_api: true

fos_elastica:
    indexes:
        theIndexName:
            types:
                theTypeName:
                    persistence:
                        driver: orm
                        model: Application\UserBundle\Entity\User
                        provider: 
                            pager_provider: true
```

The current providers are deprecated and will be removed in 1.5.x. New API will be a main thing. 

## Reasons behind PR

Right now provider does at least two things:

1. It is responsible for fetching data from the store
2. It is responsible for iterating over found objects and inserting them to elastic search. It also filters them at this step.

Because of that, there are a lot of code duplications in provider logic. It also makes it hard (not say impossible) to extend the logic. For example, I'd like to have different persistence strategies. 

1. As is it right now. A single process and iteration over the whole bunch of objects.
2. All sorts of parallel processing.

This POC splits the provider responsibility. It will be responsible only for data querying. It also introduces the `PagerPersister` object. Its responsibility is to take a pager, iterate it and persist data to ES.

The solution is completely backward compatible. To enable new features you have to call the populate cmd with `FOS_ELASTICA_USE_V5_API=1` env var or change the parameter `fos_elastica.use_v5_api` to true.

It would allow to change the persistenc logic without toucing the provider. something like this could be added to populate command:

```php
<?php
if ($input->getOption('parallel')) {
  $pagerPersister = new AsyncPagerPersister();

  $pagerPersister->insert($pager, $loggerOutput, $options);
}
```

and in the worker:

```php
<?php
/** @var ProviderV2Interface $provider */
$provider = $this->providerRegistry->getProvider($options['indexName'], $options['typeName']);
            
$pager = new Pagerfanta(new CallbackAdapter(
    function() {
        return 1;
    },
    function($offset, $length) use ($provider, $options) {
        $pager = $provider->provide($options);
        $pager->setCurrentPage($options['page']);
        
        return $pager->getCurrentPageResults();
    }
));

$this->pagerPersister->insert($pager, null, $options);
```

TODO:

* [ ] Handle disableLogging\enableLogging
* [ ] take care of batch_size option


makasim:
I'd like to continue with this but first I need an approve that something like this is likely to be merged. We can discuss the details of course. 

@lsmith77 @XWB Who is responsible for such decisions?

makasim:
Does it look good?

makasim:
@XWB could you please look at this?

XWB:
I like the idea of splitting functionality, but naming things like "v2" doesn't sound right.

makasim:
@XWB the idea was to do it BC way keeping both implementations for the current stable version. 
And in the next major release: we can remove current provider interface and rename v2 provider to just provider clean up a lot of duplications.

makasim:
I am more or less done with the design, so feel free to review and comment. I am going to work on tests now

makasim:
@lsmith77 @XWB current stable versions are 4.x, the next major is 5.x which has not been released yet. 

I guess I can do BC breaks in 5.x version, can't I?

XWB:
@makasim Yes BC breaks are allowed in 5.x (master).

XWB:
@makasim Please do not forget to re-open this PR against the master branch instead of 4.x

makasim:
@XWB the plan is to merge it to 4.x.0 than merge 4.x.0 to master (with required corrections). Then I'll remove deprecated parts in master.

BTW what do you think about this PR?

makasim:
@XWB some key parts to look at:

* [ORMPagerProvider](https://github.com/formapro-forks/FOSElasticaBundle/blob/c93bb0ad6b9383b93d68b60d3dc3387d326aaafd/Doctrine/ORMPagerProvider.php#L20)
* [InPlacePagerPersister](https://github.com/formapro-forks/FOSElasticaBundle/blob/c93bb0ad6b9383b93d68b60d3dc3387d326aaafd/Persister/InPlacePagerPersister.php#L15).
* [PopulateCommand](https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1240/files#diff-41117e4be730ae7990a8ff320ec7664cR194)

makasim:
almost done with the basics (future PRs are coming after this one). Are there any volunteers to review?

@XWB @merk @lsmith77

XWB:
This PR has been merged too soon:

1) What are the benefits of having this in 4.x ?
2) We should also squash the commits before merging as messages like "fix" are meaningless.
3) The pagination implementation remains painful and I believe https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/559 is still the right way to go.

makasim:
> 1. What are the benefits of having this in 4.x ?

Why not? It does not break anything. 

> We should also squash the commits before merging as messages like "fix" are meaningless.

What should I do now? Should I revert the merge or it is just a suggestion for future PRs? 

> The pagination implementation remains painful and I believe #559 is still the right way to go.

Why painful?

makasim:
I asked for review several times before merging. No response. It would be good if you or others at least leave a comment that you are going to review it or not.

XWB:
> Why not? It does not break anything.

Because now we have to continue supporting this feature in 4.x, which adds more work. Without it, 4.x could have kept 4.x just for security updates and continue in 5.x.

> What should I do now? Should I revert the merge or it is just a suggestion for future PRs?

It was more a suggestion for future PRs.

> Why painful?

Every time we had to add/change features to the pagination implementation, the interface needed to be changed (which cannot be done without introducing a BC break). Have a look at https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/559 and especially https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/557, it shows how pagination should be done IMO.

makasim:
>> Why not? It does not break anything.
>Because now we have to continue supporting features in 4.x, which adds more work. Without it, 4.x could have been kept just for security updates and bug fixes.

It makes it easier for developers to migrate to new provider API. I just wanted to make their life easier since it is a huge change. Of course, it added some work for me.

makasim:
> Every time we had to add/change features to the pagination implementation, the interface needed to be changed

I'll take a look

XWB:
@makasim Examples are provided in https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/689

makasim:
@XWB as far as I can see you are referring to use of paginator while searching where I am working on index population logic. I absolutely agree that pagination might not be a good choice when it comes to search.

PR #1241, opened 2017-05-08
[Propel] Restored proper object sorting

gkawka:
I've restored use of `sortingClousure` at Transformer because returned results always had random order.


XWB:
Opening one PR is enough, we can fix it in 3.x and merge the changes in 4.x and master :)

PR #1243, opened 2017-05-08
[Propel] Restored proper object sorting

gkawka:
I've restored use of sortingClousure at Transformer because returned results always had random order.


XWB:
Could you add a test as well?

gkawka:
Its quite problematic because now I can't mock db connection because Propel query is invoked with static method and don't have any idea how to test this without refactoring bunch of code.

I'll try to make some tests but I'd rather do this in other PR (will be way more code)

XWB:
:+1:

PR #1246, opened 2017-05-11
maxScore access from PaginatorAdapterInterface

istvankis:


Paginator/RawPaginatorAdapter.php
```diff
@@ -156,6 +162,19 @@ public function getSuggests()
...
+    public function getMaxScore()
+    {
+        if (!isset($this->maxScore)) {
+            $this->maxScore = $this->searchable->search($this->query)->getMaxScore();
+        }
+
+        return $this->maxScore;
+    }
+
+
```
XWB: The second empty line can be removed.

Paginator/PaginatorAdapterInterface.php
```diff
@@ -43,4 +43,10 @@ public function getAggregations();
      * @return mixed
      */
     public function getSuggests();
+
+    /**
+     * Retun max score
+     * @return float
+     */
+    public function getMaxScore();
```
XWB: Can you add this interface change to the changelog?

istvankis:
can we add this feature to 4.0.x branch?

XWB:
@istvankis As 4.0 has been released already, changing interfaces is not possible anymore.

PR #1247, opened 2017-05-11
Disable normalize keys for headers

mahho:
This snippet disables normalization (replacing - to _) for headers, for example "Content-Header" is named as "Content_Header"


XWB:
Thanks.

PR #1254, opened 2017-05-25
remove elasticsearch service from travis.yml

takeit:
Since elasticsearch is not available in trusty distro (https://docs.travis-ci.com/user/trusty-ci-environment#Data-Stores) it should not be referenced in the config at all as it is useless anyhow


XWB:
Thanks.

PR #878, opened 2015-04-30
Support for asynchronous persistence

anushr:
Added a new configuration option that defers Doctrine caused updates until after `kernel.terminate`. Useful for when fast response times are important.

PR #1261, opened 2017-06-14
Subscribe to console.terminate event when defer is true

EmmanuelVella:
This PR improves https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/878, by also subscribing to the `console.terminate` event, in order to handle entities persisted in CLI commands.


XWB:
Thanks.

PR #1278, opened 2017-08-17
Convert identifierValue to string if it is an object during deletion

spiroski:
This should address the problem of using a ValueObject as an ID. 
It will only convert the value to string if the value is object.

### DIFF
--- a/Doctrine/Listener.php
+++ b/Doctrine/Listener.php
@@ -190,7 +190,7 @@ class Listener
     private function scheduleForDeletion($object)
     {
         if ($identifierValue = $this->propertyAccessor->getValue($object, $this->config['identifier'])) {
-            $this->scheduledForDeletion[] = $identifierValue;
+            $this->scheduledForDeletion[] = is_object($identifierValue) ? (string) $identifierValue : $identifierValue;
         }
     }
 


spiroski:
Any feedback? 
I really need this, and is a a blocker for my app.

spiroski:
More then a month, and no response at all...

XWB:
I'm still not sure if this is the right solution. Implementation will fail if the object doesn't have a toString method.

spiroski:
It will fail anyway if it is an object. 
This way at least if it can be casted to string it will work.

Also in most instances, if value objects are used for IDs, they will have toString method in order to work with them ( for generating urls, for storing in db, for serialization ).

spiroski:
@XWB Any thoughts? 

I really need this, preparing to launch soon. 

This doesn't break backwards compatibility since if someone is using objects it would already be throwing an exception. There really isn't any real downside of merging this.

XWB:
How about using the [PropertyAccess Component](http://symfony.com/doc/current/components/property_access.html)?

spiroski:
How would you use it with the id? I have tried that months ago, but can't find anything that would work.

Tobion:
Fixed in #1357

PR #1281, opened 2017-09-16
Added an ssl key to the DependecyInjection config

Luciam91:
When trying to set up the `AwsAuthV4` transport to use port 443 I found this block of code in the underlying Elastica class:
https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Transport/AwsAuthV4.php#L77

I couldn't find anywhere to be able to define the `ssl` config that it was looking for, this PR adds that to a connection in the same way all of the AWS_* config is set.

If it's just a case of me missing where to set up this config, please point me in the right direction :D


XWB:
Thanks :)

PR #1283, opened 2017-09-27
Create index command

oleg-andreyev:
Right now it's impossible to create empty index with mapping without hacking things.

This PR will implement `fos:elastica:create` command for this purposes.


XWB:
Thanks :)

PR #1285, opened 2017-10-18
Improve populate command performances

wlalele:
Using the offset in the slice fetcher makes it hard
when it comes to huge volumes of data (like 100k indexes).

This modification intend to use the last item id instead of the offset.

I ran the populate command on **100K** indexes. Here are the results I got: 
- With the use of offset: **1 minute and 10 seconds**
- With this fix: **40 seconds**

We ran this fix on our pre-production environment (**1893337** indexes):
- Without the fix: **~4 hours**
- With the fix: **12 minutes and 5 seconds**

### DIFF
--- a/Doctrine/ORM/SliceFetcher.php
+++ b/Doctrine/ORM/SliceFetcher.php
@@ -22,6 +22,9 @@ use FOS\ElasticaBundle\Exception\InvalidArgumentTypeException;
  */
 class SliceFetcher implements SliceFetcherInterface
 {
+    /** @var int $lastId */
+    private $lastId = 0;
+
     /**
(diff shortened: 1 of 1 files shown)


makasim:
It assumes an id is a number, what if it is not (uuid for example)?

wlalele:
That's a good question, unfortunately I don't have a good answer 😞

makasim:
try https://github.com/php-enqueue/enqueue-elastica-bundle.

Koc:
This could be configurable and cover 80% common use cases.

makasim:
This is not applicable anymore. The provider API this PR relying on was deprecated and removed.

PR #1287, opened 2017-10-20
Remove deprecated comment in RepositoryManager.php #1275

### LINKED ISSUE #1275: RepositoryManager is deprecated
RepositoryManager is deprecated on `4.0`, but when checking the doc file for implementation for custom repository, it still shows old method.

https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/4.0.x/Resources/doc/cookbook/custom-repositories.md

What function / class / service should we replace it with?

aleixfabra:
Fix #1275


XWB:
Thanks.

PR #1290, opened 2017-10-27
Return responses in ObjectPersister methods

braimond-carelabs:


XWB:
Any reason for this change?

braimond-carelabs:
Yes, on an insert request, we cannot send back created id.

PR #1307, opened 2017-11-24
Upgrade PHP CS Fixer

keradus:
Used v1 is not maintained for over a year.

I upgraded it to v2, keeping currently configured rules as-is.

there were some violations I fixed, most of the diff is result of applying requested in config `ordered_class_elements`. If you don't feel comfortable with it, let me know - I will revert last commit.


XWB:
Looks good, but unit tests have failed.

keradus:
I don't see unit tests failing anywhere. issue is composer conflict on sf components - shall v2.8 be still supported indeed ?

XWB:
Thanks.

keradus:
you welcome ;)
actually, my question is still valid ;)

XWB:
I think we can keep s2.8 support as long as it doesn't require much effort.

keradus:
Sadly, it blocks you from update other packages, and there is already Sf not only 3, but 4 as well.

PR #1320, opened 2017-12-06
4x to master

makasim:
Contains split providers feature from PR #1240


makasim:
@XWB PR merges 4.x.0 branch into master with PR #1240. Ok to merge?

PR #1325, opened 2017-12-08
Remove deprecated stuff

makasim:


makasim:
@XWB would you like to review? PR Mostly removes previously deprecated stuff.

Command/PopulateCommand.php
```diff
@@ -210,49 +203,44 @@ private function populateIndexType(OutputInterface $output, $index, $type, $rese
...
-        } else {
-            $provider = $this->providerRegistry->getProvider($index, $type);
-            $provider->populate($loggerClosure, $options);
-        }
+        $pager = $provider->provide($options);
+
+//        $options['indexName'] = $index;
+//        $options['typeName'] = $type;
+//        $options['batch_size'] = 100;
+//        $options['skip_indexable_check'] = false;
```
Tobion: commiting commented code?

makasim: The work is still in progress, this code will be removed very soon.

composer.json
```diff
@@ -20,6 +20,7 @@
         "symfony/asset": "^2.7 || ^3.1",
         "symfony/templating": "^2.7 || ^3.1",
         "symfony/translation": "^2.7 || ^3.1",
+        "pagerfanta/pagerfanta": "^1.0.5",
```
Tobion: this is not a hard dependency at all. we don't use the persistence stuff is optional and we don't use it for example. so we also do not need this pagination, do we?

makasim: Since 5.x it is used by default. Here:

* https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Doctrine/MongoDBPagerProvider.php
* https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Doctrine/ORMPagerProvider.php
* https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/Provider/PagerfantaPager.php

Though FOSElasticaBundle relies on its own `PagerInterface` so one can simply use any other pager implementations or their own.

Tobion: It is only used if the persistence configuration of a type is enabled. So it's optional.

PR #1302, opened 2017-11-17
Adding Symfony4 support

geecu:
#SymfonyConHackday2017

PR #1326, opened 2017-12-08
remove unused dependencies

Tobion:
@XWB or @makasim please merge this first, then I can look into finishing #1302

### DIFF
--- a/Tests/Functional/WebTestCase.php
+++ b/Tests/Functional/WebTestCase.php
@@ -20,9 +20,9 @@
 
 namespace FOS\ElasticaBundle\Tests\Functional;
 
-use Symfony\Bundle\FrameworkBundle\Tests\Functional\WebTestCase as BaseWebTestCase;
+use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase as BaseKernelTestCase;
 
-class WebTestCase extends BaseWebTestCase
+class WebTestCase extends BaseKernelTestCase
 {
     protected static function getKernelClass()
     {
--- a/Tests/Functional/WebTestCase.php
+++ b/Tests/Functional/WebTestCase.php
@@ -21,7 +21,11 @@
 namespace FOS\ElasticaBundle\Tests\Functional;
 
 use Symfony\Bundle\FrameworkBundle\Test\KernelTestCase as BaseKernelTestCase;
+use Symfony\Component\Filesystem\Filesystem;
 
+/*
+ * Based on https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bundle/FrameworkBundle/Tests/Functional/WebTestCase.php
+ */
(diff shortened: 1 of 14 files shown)


Tests/Functional/app/Basic/config.yml
```diff
@@ -1,10 +1,6 @@
 imports:
     - { resource: ./../config/config.yml }
 
-framework:
-    templating:
```
Tobion: templating is not needed, twig is enough for the webprofiler integration

Tests/Functional/app/Basic/bundles.php
```diff
@@ -10,13 +10,11 @@
...
 
 use FOS\ElasticaBundle\FOSElasticaBundle;
-use Knp\Bundle\PaginatorBundle\KnpPaginatorBundle;
 use Symfony\Bundle\FrameworkBundle\FrameworkBundle;
 use Symfony\Bundle\TwigBundle\TwigBundle;
 
 return [
     new FrameworkBundle(),
     new FOSElasticaBundle(),
-    new KnpPaginatorBundle(),
```
Tobion: there is no functional test that uses this (there is PaginateElasticaQuerySubscriberTest which is a unit test and only needs knplabs/knp-components)

composer.json
```diff
@@ -17,9 +17,6 @@
         "symfony/console": "^2.7 || ^3.1",
         "symfony/options-resolver" : "^2.7 || ^3.1",
         "symfony/property-access": "^2.7 || ^3.1",
-        "symfony/asset": "^2.7 || ^3.1",
-        "symfony/templating": "^2.7 || ^3.1",
-        "symfony/translation": "^2.7 || ^3.1",
```
Tobion: those things are not needed at all, it was just added because the functional test enabled those components for no reason, see Tests/Functional/app/config/config.yml

composer.json
```diff
@@ -17,9 +17,6 @@
         "symfony/console": "^2.7 || ^3.1",
         "symfony/options-resolver" : "^2.7 || ^3.1",
         "symfony/property-access": "^2.7 || ^3.1",
-        "symfony/asset": "^2.7 || ^3.1",
```
Tobion: not needed since !742

Tests/Functional/WebTestCase.php
```diff
@@ -20,9 +20,9 @@
 
 namespace FOS\ElasticaBundle\Tests\Functional;
 
-use Symfony\Bundle\FrameworkBundle\Tests\Functional\WebTestCase as BaseWebTestCase;
```
Tobion: this was using the wrong namespace (Tests vs Test) and also no need to use WebTestCase if you just need the container

XWB:
There seems to be a merge conflict.

PR #1327, opened 2017-12-09
Cast identifier to string in Transformer and Doctrine listener

vicdelfant:
While #1067 solved issues when using value objects as the ID for the `AbstractElasticaToModelTransformer`, the `ModelToElasticaAutoTransformer` has a similar issue when an object is used as the model's ID.

This results in an object as the document's `_id` parameter instead of its value, which then causes the following when populating indexes:
```
Malformed action/metadata line [1], expected a simple value for field [_id] but found [START_OBJECT]
```

This PR fixes this behavior by casting the model's identifier to a string. The `Document` then transforms correctly to:

```
array:1 [
  0 => Elastica\Document {#962
    #_data: array:1 [
      "firstname" => "Vic"
    ]
    #_docAsUpsert: false
    #_autoPopulate: false
    #_upsert: null
    #_params: array:3 [
      "_id" => "00000000-0000-0000-0000-000000000000"
      "_type" => ""
      "_index" => ""
    ]
    #_rawParams: []
  }
]
```

Unit test is included.

### DIFF
--- a/Doctrine/Listener.php
+++ b/Doctrine/Listener.php
@@ -222,7 +222,12 @@ class Listener
      */
     private function scheduleForDeletion($object)
     {
-        if ($identifierValue = $this->propertyAccessor->getValue($object, $this->config['identifier'])) {
+        $identifierValue = $this->propertyAccessor->getValue($object, $this->config['identifier']);
+        if ($identifierValue && !is_scalar($identifierValue)) {
+            $identifierValue = (string) $identifierValue;
+        }
+
+        if ($identifierValue) {
             $this->scheduledForDeletion[] = $identifierValue;
         }
     }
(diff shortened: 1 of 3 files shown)


vicdelfant:
There were two tests using an `assertSame` on the document ID which forced the return value to be an integer, so I added a typecast to fix this. We could of course also use `assertEquals` instead.

I assume that ID's aren't expected to be integers, at least not with `Document` using empty strings as the placeholder value.

vicdelfant:
I have improved the earlier changes so that the `(string)` cast is only applied to non-scalar identifiers. Tests have been reverted to their original behavior as well, so we're back to `$this->assertSame()` like it should.

Additionally, the Doctrine listener now also properly casts non-scalar identifiers when deleting entities.

Tobion:
Merged in #1357. Thank you @vicdelfant

vicdelfant:
Thanks for merging @Tobion! 👍

PR #1335, opened 2017-12-12
Use assertEmpty and assertNull

carusogabriel:


makasim:
Thank you for the fix.

PR #1337, opened 2017-12-12
[travis] Use dunglas/symfony-lock to fix installed Symfony component versions.

makasim:


makasim:
Raises minimum supported Symfony version to 2.8 and 3.4 (LTS releases)

PR #1339, opened 2017-12-12
Unused dependencies

makasim:
same as https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1326 but it fixes a conflict


makasim:
@XWB same PR with a fixed confilct

makasim:
That was great @Tobion. Lots of cleanups! I like it.

PR #1345, opened 2017-12-14
Call insert method from pagePersister  instead of from the registry

nicolassing:


Command/PopulateCommand.php
```diff
@@ -265,7 +266,7 @@ function(PostAsyncInsertObjectsEvent $event) use ($loggerClosure) {
         $options['indexName'] = $index;
         $options['typeName'] = $type;
 
-        $this->pagerPersisterRegistry->insert($pager, $options);
```
makasim: this should be `$this->pagerPersister->insert($pager, $options);`

makasim: pagerPersister is set in initialize method

wenmingtang: 🆒

PR #1352, opened 2017-12-15
split directory structure in doc, src and test

Tobion:
- for better clarity and better autoloading by excluding tests
- also with .gitattributes for optimized dist packages
- having docs and license in Resources/ is not recommended anymore

Issue #801, opened 2015-02-25
Seeking Propel maintainer


Lionel09:
I'm using Propel1 with this bundle and have troubles trying to use the finder searching the full index.

This piece of code in the `transform` method is incompatible with the PropelCollection:

```
            $transformed[$type] = array_combine(
                array_map(
                    function($o) use ($identifierGetter) {
                        return $o->$identifierGetter();
                    },
                    $transformedObjects
                ),
                $transformedObjects
            );
```

If I replace with this code, it's ok:

```
            foreach($transformedObjects as $object) {
                $transformed[$type][$object->$identifierGetter()] = $object;    
            }
```

No time for PR, but just want to tell it if someone is interested!

merk:
Propel support is no longer maintained. Unless someone using it provides PRs for updates it will remain as is.

StudioEcho:
OK, so if I have some time, I'll try to make some PRs to keep Propel support compatible with this bundle.

Tobion:
Closing due to #1354

PR #1354, opened 2017-12-16
remove Propel support

Tobion:
- Propel1 is unmaintained since many years
- FOSElasticaBundle did not mantain propel either according to https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/801#issuecomment-75946976
- FosUserBundle basically dropped propel as well

Let's get rid of it in the new major version.

- Fix phpunit paths after #1352


UPGRADE-5.0.md
```diff
@@ -26,10 +29,7 @@ Removed stuff:
 * `FOS\ElasticaBundle\Doctrine\ORM\Provider` class was removed.
 * `FOS\ElasticaBundle\Doctrine\PHPCR\SliceFetcher` class was removed.
 * `FOS\ElasticaBundle\Doctrine\PHPCR\Provider` class was removed.
-* `FOS\ElasticaBundle\Propel\Provider` class was removed.
```
makasim: This has to be reverted

makasim:
👍

UPGRADE-5.0.md
Tobion: I added that propel support was removed. This class is only a subset of propel classes. So I don't see the value in only mentioning this class.

PR #1359, opened 2017-12-18
remove deprecated offset and batch-size populate options

Tobion:


src/Command/PopulateCommand.php
```diff
@@ -220,35 +213,33 @@ private function populateIndexType(OutputInterface $output, $index, $type, $rese
...
-            );
+        $this->dispatcher->addListener(
+            Events::POST_INSERT_OBJECTS,
+            function(PostInsertObjectsEvent $event) use ($loggerClosure) {
+                $loggerClosure(count($event->getObjects()), $event->getPager()->getNbResults());
+            }
+        );
 
-            $this->dispatcher->addListener(
-                Events::POST_ASYNC_INSERT_OBJECTS,
```
Tobion: @makasim that event does not seem to be dispatched at all. What's the idea of that?

makasim: It should be used by async pager persister. It does not deal with objects directly but schedule sub jobs to some workers. Like here https://github.com/php-enqueue/enqueue-elastica-bundle/blob/master/Persister/QueuePagerPersister.php#L125

Tobion:
Good to merge?

PR #1362, opened 2017-12-19
Fix typo in populate command: option first-page

bujnos:


src/Command/PopulateCommand.php
```diff
@@ -139,7 +139,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $options['last_page'] = $input->getOption('last-page');
         }
 
-        if ($input->isInteractive() && $reset && 1 < $input->getOption('first_page')) {
+        if ($input->isInteractive() && $reset && 1 < $input->getOption('first-page')) {
```
clement-michelet: You stored the value of `$input->getOption('first-page')` in `$options['first_page']`. 
Why not re-use that value ?

bujnos: A good point, I changed it.

XWB:
Thanks.

trsteel88:
I think we need a new 4.1.1 tag for this?

Issue #1364, opened 2017-12-19
Exception: The options "first_page", "max_per_page" do not exist.


spiroski:
I get the following error when I try to execute the populate command on the latest version of the 4.x branch.

```
[Symfony\Component\OptionsResolver\Exception\UndefinedOptionsException]
  The options "first_page", "max_per_page" do not exist. Defined options ar
  e: "batch_size", "clear_object_manager", "debug_logging", "ignore_errors"
  , "indexName", "offset", "pager_provider", "query_builder_method", "reset
  ", "skip_indexable_check", "sleep", "typeName".
```

XWB:
Should be fixed by https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1362

spiroski:
The fix is not applied to the 4.x branch.

XWB:
Can you try out https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1365 ?

spiroski:
Yes, that fixed the problem.

PR #1365, opened 2017-12-20
fix options does not exist.

makasim:
fixes https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1364


XWB:
Shouldn't it be the same fix as https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1362 ?

makasim:
As far as I can see

makasim:
@XWB  In any case, those options do not make any sense for legacy provider API.

Issue #1369, opened 2017-12-20
Upgrade from 3.2.1 to 5


valimero:
Hi everybody,

I have a problem when I do `phpc cache:clear --no-warmup`
![image](https://user-images.githubusercontent.com/15141970/34211395-3aaa75ec-e599-11e7-9639-1cf540efe9b6.png)

Do you know where it can come from ?

valimero:
Is there a problem in /src/Resources/config/mongodb.xml:11 ?

``` xml
        <service id="fos_elastica.pager_provider.prototype.mongodb" class="FOS\ElasticaBundle\Doctrine\MongoDBPagerProvider" public="true" abstract="true">
            <argument type="service" id="doctrine" /> <!-- HERE --> 
            <argument type="service" id="fos_elastica.doctrine.register_listeners" />
            <argument /> <!-- model -->
            <argument type="collection" /> <!-- options -->
        </service>
```

PR #1370, opened 2017-12-20
Use correct Doctrine registry

makasim:
fixes https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1369

Should be merged to 5.x too.


franmomu:
Hi, it would be nice to merge it in 5.x soon because it doesn't work with mongodb right now.

valimero:
You can put temporary “dev-master” in your composer.json !

franmomu:
I know, thanks! I was a reminder, just in case :)

jacquelotjeff:
Hello, 
I'm agree with franmomu, it's a real bug not an improvement ... 
It should be merge now in 5.x tags because many people are blocked (and using dev-master tag is not a good practice !)

Thanks.

Issue #1375, opened 2017-12-24
fos_elastica.pager_provider service for mongodb is not well configured


jacquelotjeff:
Hello,

I use FOSElasticaBundle in 5.0 tags with mongodb.

I have configured my type in the configuration with mongodb driver : 
```yaml
           types:
                user:
                    persistence:
                        driver: mongodb
```

But I have this error : 
```
    [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]     
    The service "fos_elastica.pager_provider.app.user" has a dependency on     
    a non-existent service "doctrine". 
```

Of course the service Doctrine doesn't exist because I use mongodb. I think it is rather the doctrine_mongodb service that should be used here.

I think the error is in the mongodb.xml file (located : src/Resources/config/mongodb.xml) in line 10 : 

```xml
        <service id="fos_elastica.pager_provider.prototype.mongodb" class="FOS\ElasticaBundle\Doctrine\MongoDBPagerProvider" public="true" abstract="true">
            <argument type="service" id="doctrine" /> <!-- SHOULD BE doctrine_mongodb-->
            <argument type="service" id="fos_elastica.doctrine.register_listeners" />
            <argument /> <!-- model -->
            <argument type="collection" /> <!-- options -->
        </service>
```

I can make a pull request to change this if you want.

Thank you for your feedback.
Jeff

Tobion:
#1370

PR #1376, opened 2017-12-24
Replace mongodb reqs to doctrine_mongodb for pager service

jacquelotjeff:
This PR patch #1375 issue on 5.* tag.

More info in #1375.

```
    [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]     
    The service "fos_elastica.pager_provider.app.user" has a dependency on     
    a non-existent service "doctrine". 
```


Tobion:
Thanks but this has already been fixed in #1370

PR #1380, opened 2018-01-02
Change minimum php version

lchrusciel:
Small fix to be more consistent with semantic versioning. We cannot be 100% sure, that this bundle will work with php 8


XWB:
Thanks.

lchrusciel:
You're welcome :)

PR #1390, opened 2018-01-31
Added usage documentation on how to pass general ES query parameters.

manuelkiessling:
I had a hard time finding out exactly where and how to pass ES query parameters - in my case, I needed to have `?search_type=dfs_query_then_fetch` added to the ES HTTP call, but couldn't find any hints in the docs.

I have added a paragraph to the usage chapter which should point people into the right direction.


XWB:
Thanks.

PR #1395, opened 2018-02-08
Add proper types for the TransformEvent

stof:
Types are not mixed on these variables

### DIFF
--- a/src/Event/TransformEvent.php
+++ b/src/Event/TransformEvent.php
@@ -20,6 +20,7 @@
 
 namespace FOS\ElasticaBundle\Event;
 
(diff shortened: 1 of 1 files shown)


src/Event/TransformEvent.php
```diff
@@ -78,15 +79,15 @@ public function getFields()
...
     public function getObject()
     {
         return $this->object;
     }
 
     /**
-     * @param mixed $document
+     * @param Document $document
      */
     public function setDocument($document)
```
stof: Note that I have not added the typehint here, as it would be a BC break (method is not final)

PR #1396, opened 2018-02-08
Remove bad duplicate license headers

stof:
All files have the right header (and PHP-CS-Fixer is configured to enforce it). But a bundle of files were holding a second header after it.


XWB:
Apparently this mistake was made after upgrading to PHP-CS-Fixer 2.

PR #1397, opened 2018-02-08
Add autowiring support for Elastica\Client

stof:
The extending client provided by the bundle to add extra logging should not be enforced for typehints. It creates extra coupling to the bundle. The typehint should rather be on the library class.


stof:
I kept the autowiring alias for `FOS\ElasticaBundle\Elastica\Client` though, as removing it would be a BC break.

PR #1402, opened 2018-02-21
Fix pagination of ORM queries when populating

stof:
Paginating a DQL query which does not have a ORDER BY clause does not work cross-platform, as the sorting is undefined in such case.
MySQL is not affected by the bug because it always adds an implicit sorting by insertion order (which is a compliant way to implement the undefined order but tends to hide this issue from developers having to deal with other platforms too). But PostgreSQL is impacted for instance.

### DIFF
--- a/src/Doctrine/ORMPagerProvider.php
+++ b/src/Doctrine/ORMPagerProvider.php
@@ -12,6 +12,8 @@
 namespace FOS\ElasticaBundle\Doctrine;
 
 use Doctrine\Common\Persistence\ManagerRegistry;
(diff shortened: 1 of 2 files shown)


stof:
Note that unit tests are not covering this new logic for now. But as there are mocking the QueryBuilder currently, they would not even ensure that the logic produces the expected result (as they would test that they do it fine according to the behavior of the mock, with no guarantee it is actually the Doctrine behavior).
Testing this in a reliable way would require an integration test using an actual QueryBuilder.

PR #1406, opened 2018-03-07
Array check

jongotlin:
`$data` can be either array or string https://github.com/ruflin/Elastica/blob/6.0.1/lib/Elastica/Client.php#L673


jongotlin:
Rebased but please don't merge yet.

jongotlin:
Ready

XWB:
Thanks.

PR #1427, opened 2018-06-12
Update Pagerfanta to 2.0.0

bitbager:


bitbager:
We need it dramatically for our https://github.com/BitBagCommerce/SyliusElasticsearchPlugin. Sylius currently uses v2.0.0 Pagerfanta, which disables us from using our plugin with it and FOS Elastica Bundle as well.

XWB:
This needs to be changed to `^1.0.5|^2.0` as this bundle still supports PHP 5.6.

bitbager:
Done 👌

bitbager:
Thank you! 🙂

PR #1442, opened 2018-08-09
Allow setting the refresh option for persisters

jvasseur:
This allow using the [refresh](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html) option in persisters.


XWB:
Thanks.

PR #1454, opened 2018-09-19
Fix a defect when populating a new type in 3.2.x

edhgoose:
We've spotted a problem where if we try and reset a type using:

`php app/console fos:elastica:reset --index=index_name --type=type_name -v `

... we get an error saying:

```
ActionRequestValidationException[Validation Failed: 1: no settings to update;]
```

Our understanding of this is that the `$settings` array starts out containing the `number_of_shards` and `number_of replicas`, but these are then removed. This leaves the `$settings` array empty again, and ElasticSearch isn't happy with an empty `$settings` array.

We've applied the following and seen the issue resolved.


XWB:
Closing as 3.2.x is not maintained anymore.

PR #1469, opened 2018-11-18
Document how to use AWS SQS, with FosElasticaBundle and enqueue

jandom:
This was never documented end-to-end with such detail before


XWB:
Thank you for the effort 👍

PR #1475, opened 2018-12-16
Fix deprecation for symfony/config 4.2+

franmomu:
This is just a fix to avoid [the deprecation warning](https://symfony.com/blog/new-in-symfony-4-2-important-deprecations#deprecated-tree-builders-without-root-nodes) when use symfony/config 4.2+


XWB:
Thanks.

PR #1481, opened 2019-01-14
Put a random return in README.md to demonstrate build breaks

jandom:
This PR verifies that CI is broken for master, due to a dependency change

A random return is introduced into the README.md file, yet the build will probably break

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,8 @@
 FOSElasticaBundle
 =================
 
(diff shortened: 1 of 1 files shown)


jandom:
Resolved by https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1482

Thanks @XWB !

PR #1483, opened 2019-01-15
Fix deprecation for symfony/config 4.2+

franmomu:
I missed all the other uses of `TreeBuilder` in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1475 🤦‍♂️

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -77,8 +77,7 @@ class Configuration implements ConfigurationInterface
      */
     public function getDynamicTemplateNode()
     {
-        $builder = new TreeBuilder();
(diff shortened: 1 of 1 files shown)


jandom:
This piece of code seems to be repeated over and over again 

```
        if (method_exists($builder, 'getRootNode')) {
             $node = $builder->getRootNode();
         } else {
             // BC layer for symfony/config 4.1 and older
             $node = $builder->root('types');
         }
```

Would it make sense to pull it out into a function, or somehow reuse to provide the BC layer?

XWB:
I agree with @jandom

XWB:
Thanks :)

PR #1488, opened 2019-01-19
Testing AbstractElasticaToModelTransformer

jandom:
Remove getSortingClosure, as it is not used anywhere
Introduce a UnitTestHelper class for some shared functionality


jandom:
@XWB hi there – anything sticky with this one that i could factor out to make the PR merged faster?

Issue #327, opened 2013-06-25
Annotations at Entities to add configurations for DoctrineORM


ravindrakhokharia:
While using FOSElasticaBundle, i have found that i need to include many entities for elastic search, So i was wondering if there is any possibility i can auto-map entities or may be there are Annotations available at entities, using which i can include the configurations like mapping, boosts, etc for specific entity.

It would be good if there is some functionality for configuring elastic mapping at entity class.

PR #1490, opened 2019-01-22
clarify the features in the readme file

dbu:
we got quite confused by the claim that this bundle would generate mappings - from how i understand https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/327#issuecomment-47928727 and other issues here, and from what i found in the code, the bundle does never generate mappings. it "only" allows to let elasticsearch do dynamic mapping, and it allows to configure an explicit mapping.

is that correct or do i miss something?


XWB:
You are right @dbu :+1:

PR #1496, opened 2019-02-04
Allow setting transport connection persistency

damijank:
ES services, especially the managed ones (like Azure's), tend to break a connection if it's idle long enough. This PR introduces a configuration setting to be propagated through to the transport layer setup, allowing to use a non-persistent connection (the implicit default there is a persistent connection).

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -488,6 +488,7 @@ class Configuration implements ConfigurationInterface
                                         ->scalarNode('retryOnConflict')
                                             ->defaultValue(0)
                                         ->end()
+                                        ->booleanNode('persistent')->defaultValue(true)->end()
                                     ->end()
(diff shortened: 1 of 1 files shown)


XWB:
Thanks :)

damijank:
No problem @XWB 

What are your plans for a new release? Do these 60+ new commits on master contain any breaking changes in regards to 5.0.3 (the latest release)?

XWB:
@damijank The next release should be 6.0 as there are some BC breaks in master. This will also allow us to refactor the configuration (for example, multiple types within an index are deprecated since Elasticsearch 6).

damijank:
@XWB too bad :disappointed: So for the time being we have to (and already have) override this... we're using this in a long-running process (a RMQ worker) and with a persistent connection every so often this raises an exception.
We'll be looking out for the new release. Thank you!

PR #1519, opened 2019-05-13
Check that element have at least one value

biozshock:
According to https://github.com/symfony/symfony/issues/20356

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -458,7 +458,7 @@ class Configuration implements ConfigurationInterface
                                                 ->ifTrue(function ($v) { return !is_array($v); })
                                                 ->then(function ($v) { return array($v); })
                                             ->end()
-                                            ->cannotBeEmpty()
+                                            ->requiresAtLeastOneElement()
                                             ->defaultValue([400, 403, 404])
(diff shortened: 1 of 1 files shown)


killerwolf:
`Using Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition::cannotBeEmpty() at path "fos_elastica.clients..connections..http_error_codes" has no effect, consider requiresAtLeastOneElement() instead. In 4.0 both methods will behave the same.`

XWB:
Thanks :)

biozshock:
@XWB may we have a new tag?

blackarcanis:
Not sure to understand how we can avoid this depreciation message...
I'm using the 5.1 version of ElasticaBundle on last 3.4 Symfony version and I would like to not have this depreciation message anymore. Anything I can do ?

biozshock:
@blackarcanis Easy! This PR is not yet have a tag. Latest version 5.1.0 does not have it yet: https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/v5.1.0/src/DependencyInjection/Configuration.php#L461
Just wait till new tag is released.

blackarcanis:
@biozshock OK great, thank you ! :)

PR #1523, opened 2019-05-28
[documentation] missing content type to req

cbastienbaron:
missing content type to req else error 406 `{"error":"Content-Type header [application/x-www-form-urlencoded] is not supported","status":406}`


XWB:
Thanks :)

Issue #1530, opened 2019-06-12
Issues with and the dispatch() method in Symfony 4.3 when using EnqueueBundle


boroth:
I'm trying to use the EnqueueBundle with FOSElasticaBundle to speed up the populate command for about 6 million Entities.

I'm on Symfony 4.3, which seems to be the main issue. Are there any other known issues for FOSElasticaBundle with Symfony 4.3? 

Editing with the "solution" I found:
It seems like the difference between `Symfony\Component\EventDispatcher\Event` and `Symfony\Contracts\EventDispatcher\Event` happened between [Symfony 3](https://symfony.com/doc/3.4/components/event_dispatcher.html) and [Symfony 4](https://symfony.com/doc/current/components/event_dispatcher.html).

The change in the dispatch() method was introduced in Symfony 4.3. Here is the [documentation](https://symfony.com/doc/4.2/components/event_dispatcher.html#dispatch-the-event) for [dispatch()](https://github.com/symfony/symfony/blob/4.2/src/Symfony/Component/EventDispatcher/EventDispatcher.php#L37) in version 4.2, and here is the [documentation](https://symfony.com/doc/current/components/event_dispatcher.html#dispatch-the-event) for [dispatch()](https://github.com/symfony/symfony/blob/4.3/src/Symfony/Component/EventDispatcher/EventDispatcher.php#L51) in 4.3

Since creating this issue, I've submitted a [Pull Request](https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1531).

I've gotten the EnqueueBundle setup, and as far as I can tell, the consumer starts running correctly, but immediately starts constantly outputting the same error repeatedly. This is before even running the `fos:elastica:populate` command. So, when I run:

```
./bin/console enqueue:consume --setup-broker -vvv
```

I first receive this output:
```
  [app] [FsDriver] Declare router queue "enqueue.app.default" file: /SymfonyDir/var/queue/enqueue.app.default
21:21:17 DEBUG     [app] [FsDriver] Declare processor queue "fos_elastica_populate" file: /SymfonyDir/var/queue/fos_elastica_populate
21:21:17 DEBUG     [app] [ExclusiveCommandExtension] This is a exclusive command queue and client's properties are not set. Setting them
```

After about 3 or 4 seconds, I keep getting the following output:
```
[...]

PR #1531, opened 2019-06-13
Updates for Events in Symfony 4.3 (Issue #1530)

boroth:
This pull request is regarding the issue in #1530.

Updates Symfony\Component\EventDispatcher\Event to Symfony\Contracts\EventDispatcher\Event
Updates dispatch() method to swap the order of the parameters.

### DIFF
--- a/src/Command/PopulateCommand.php
+++ b/src/Command/PopulateCommand.php
@@ -177,7 +177,7 @@ class PopulateCommand extends Command
     private function populateIndex(OutputInterface $output, $index, $reset, $options)
     {
         $event = new IndexPopulateEvent($index, $reset, $options);
-        $this->dispatcher->dispatch(IndexPopulateEvent::PRE_INDEX_POPULATE, $event);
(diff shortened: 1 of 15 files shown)


boroth:
Whoops, I guess I need to fix the tests as well.

manu-sparheld:
We need it, can someone merge it ?

giovannialbero1992:
@boroth If I remember, You should start an ES instance reachable by localhost on port 9200, you could use docker and launch phpunit wiht ES active, otherwise you can switch on travis on your fork and see tests result but this method is sure more slow than local tests.
Let me know if it is ok 😄

j-mywheels:
This breaks BC for Symfony 3.4^ & 4.3 <. Still needs to be supported tbh. You need to use the LegacyEventDispatcherProxy class to provide forward and backward compatibility.

See: https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching#supporting-both-dispatchers @boroth @cbastienbaron

j-mywheels:
WIP: https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1537

XWB:

PR #1542, opened 2019-07-17
reorder dispatch() arguments, #1537

tacman:


src/Command/PopulateCommand.php
```diff
@@ -183,7 +183,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
     private function populateIndex(OutputInterface $output, $index, $reset, $options)
     {
         $event = new IndexPopulateEvent($index, $reset, $options);
-        $this->dispatcher->dispatch(IndexPopulateEvent::PRE_INDEX_POPULATE, $event);
+        $this->dispatcher->dispatch($event, IndexPopulateEvent::PRE_INDEX_POPULATE);
```
maxhelias: To avoid break change, you should use this on every `dispatcher` : 

```suggestion
        if (class_exists(LegacyEventDispatcherProxy::class)) {
                $this->dispatcher->dispatch($event, IndexPopulateEvent::PRE_INDEX_POPULATE);
        } else {
                $this->dispatcher->dispatch(IndexPopulateEvent::PRE_INDEX_POPULATE, $event);
        }
```

Zizitto: When this change will be released?. I have huge count of logs related to this issue.

XWB:

PR #1543, opened 2019-07-17
re-order ->dispatch() arguments, #1537

tacman:


tacman:
oh, drat, looks like tests are failing, since they are explicitly looking for the arguments in the old order.

XWB:

PR #1552, opened 2019-08-29
missed default config section

zaartix:
There is no recipe to create "enqueue_elastica" config. Without this config script is crushing with error "The child node "transport" at path "enqueue_elastica" must be configured."


XWB:
:+1:

PR #1553, opened 2019-09-02
Change callback argument to null in Client

franmomu:
I was trying the `master` branch of `elastica` and [the `$callback` argument of ConnectionPool](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Connection/ConnectionPool.php#L38) has now a `callable` type, this `$callback` comes from [Client](https://github.com/ruflin/Elastica/blob/master/lib/Elastica/Client.php#L67).

The thing is that apparently, defining `<argument />` means an empty string, so it fails to create the Client.


franmomu:
I've just realised that https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1551 is for the same thing 🤦‍♂

XWB:
^^

PR #1562, opened 2019-10-08
ResetTemplates and Populate commands improvements.

ekyna:
Do not prompt for confirmation if not interactive. 
New line after index refresh.

### DIFF
--- a/src/Command/ResetTemplatesCommand.php
+++ b/src/Command/ResetTemplatesCommand.php
@@ -68,7 +68,7 @@ final class ResetTemplatesCommand extends Command
         $indexTemplate = $input->getOption('index');
         $deleteByPattern = $input->getOption('force-delete');
 
-        if ($deleteByPattern) {
+        if ($input->isInteractive() && $deleteByPattern) {
             $helper = $this->getHelper('question');
             $question = new ConfirmationQuestion('You are going to remove all template indexes. Are you sure?', false);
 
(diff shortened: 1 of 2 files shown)


src/Command/ResetTemplatesCommand.php
```diff
@@ -85,6 +85,6 @@ protected function execute(InputInterface $input, OutputInterface $output)
             $this->resetter->resetAllIndexes($deleteByPattern);
         }
 
-        return 0;
+        return;
```
XWB: This change is not good practice and will be deprecated from Symfony 4.4: https://github.com/symfony/symfony/blob/4.4/UPGRADE-4.4.md

src/Command/ResetTemplatesCommand.php
```diff
@@ -68,12 +68,12 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         $deleteByPattern = $input->getOption('force-delete');
 
-        if ($deleteByPattern) {
+        if ($input->isInteractive() && $deleteByPattern) {
             $helper = $this->getHelper('question');
             $question = new ConfirmationQuestion('You are going to remove all template indexes. Are you sure?', false);
 
             if (!$helper->ask($input, $output, $question)) {
-                return 1;
+                return;
```
XWB: Same as below.

src/Command/ResetTemplatesCommand.php
ekyna: Reverted.

src/Command/ResetTemplatesCommand.php
ekyna: Reverted.

XWB:
:+1:

PR #1566, opened 2019-10-24
change dispatch() calls to symfony 4.3 syntax

chypriote:
Following the discussion in #1537, I've updated the calls to dispatch().


stloyd:
This is no-go in forcing new format of call as we have to support older versions (LTS) which don't support that way.

This can be done only if you add some proxy function that will check version of Symfony & then decide about format of call.

chypriote:
That's the point of the LegacyEventDispatcherProxy as seen [in symfony's doc](https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching#supporting-both-dispatchers)

XWB:
But the `LegacyEventDispatcherProxy` class does not exists in older Symfony versions.

chypriote:
Oh yeah you are right

yunosh:
Any plans to merge this PR?

XWB:

PR #1567, opened 2019-10-25
Added a shortcut to return a pager for hybrid results

COil:
This is a method I have added in one of my project. It would be nice to have it in the bundle. 
Sorry, I made this quite rapidly. What do I need to add to the PR?

### DIFF
--- a/src/Finder/TransformedFinder.php
+++ b/src/Finder/TransformedFinder.php
@@ -74,7 +74,7 @@ class TransformedFinder implements PaginatedFinderInterface
      */
     public function findPaginated($query, $options = [])
     {
-        $paginatorAdapter = $this->createPaginatorAdapter($queryObject, $options);
+        $paginatorAdapter = $this->createPaginatorAdapter($query, $options);
 
         return new Pagerfanta(new FantaPaginatorAdapter($paginatorAdapter));
     }
--- a/src/Finder/TransformedFinder.php
+++ b/src/Finder/TransformedFinder.php
@@ -74,7 +74,6 @@ class TransformedFinder implements PaginatedFinderInterface
      */
     public function findPaginated($query, $options = [])
     {
-        $queryObject = Query::create($query);
         $paginatorAdapter = $this->createPaginatorAdapter($queryObject, $options);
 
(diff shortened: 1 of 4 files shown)


COil:
Hi @XWB , should the method be added in the interface?

COil:
> Hi @XWB , should the method be added in the interface?

I have added the method in the interface.

COil:
You can see how it can be used here: https://www.strangebuzz.com/en/blog/implementing-a-search-engine-with-elasticsearch-and-symfony-part-2?q=elasticsearch#h2_5

COil:
Hi @lsmith77 , can it be merged please?

src/Finder/TransformedFinder.php
```diff
@@ -80,6 +80,16 @@ public function findPaginated($query, $options = [])
         return new Pagerfanta(new FantaPaginatorAdapter($paginatorAdapter));
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function findHybridPaginated($query)
+    {
+        $paginatorAdapter = $this->createHybridPaginatorAdapter($query);
```
lsmith77: don’t we need some code here to ensure we have a query instance?

COil: Hi @lsmith77  , it's done in `createHybridPaginatorAdapter` -> `$query = Query::create($query);` -> `throw new InvalidException('Unexpected argument to create a query for.');`

lsmith77: ah thx.

COil: > I guess this is convenient syntax sugar?

Yes. It avoid me to add this function at two places in a project : 
https://www.strangebuzz.com/en/blog/implementing-a-search-engine-with-elasticsearch-and-symfony-part-2 
Check out the commented line.

> shouldn’t we then also one such method using createRawPaginatorAdapter()?

I am not sure to understand what you mean.

lsmith77: ie. adding a `findRawPaginated()`

COil: > ah thx.

The `$queryObject = Query::create($query);` call in the `findPaginated` function seems useless as it's already done in `createPaginatorAdapter` function.

PR #1574, opened 2019-11-27
Fix PaginateElasticaQuerySubscriberTest

jschaedl:
This will fix:

```
FOS\ElasticaBundle\Tests\Unit\Subscriber\PaginateElasticaQuerySubscriberTest::testShouldDoNothingIfSortParamIsEmpty
TypeError: Argument 1 passed to Symfony\Component\HttpFoundation\Request::get() must be of the type string, null given, called in ./FOSElasticaBundle/src/Subscriber/PaginateElasticaQuerySubscriber.php on line 153
```


XWB:
👍

PR #1575, opened 2019-11-27
Add return codes to commands to fix deprecated null return values

jschaedl:
Non-int status codes in Command::execute() are deprecated since Symfony 4.4:

- https://github.com/symfony/symfony/issues/33747
- https://github.com/symfony/symfony/pull/33775


XWB:
👍

PR #1577, opened 2019-11-28
Add alias to index on index creation

jschaedl:
I have configured aliases for my indexes with `use_alias: true` and wanted to use . `./bin/console fos:elastica:create` to create them. 
I expected the command to create the index and adding the configured alias, but it only created the indexes. So here is a fix for that.

### DIFF
--- a/src/Command/CreateCommand.php
+++ b/src/Command/CreateCommand.php
@@ -70,6 +70,10 @@ class CreateCommand extends Command
             }
             $mapping = $this->mappingBuilder->buildIndexMapping($indexConfig);
             $index->create($mapping, false);
+
+            if ($indexConfig->isUseAlias()) {
+                $index->addAlias($indexName);
+            }
         }
 
         return 0;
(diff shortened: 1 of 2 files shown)


COil:
Hi, you must use the "use_alias: true" option for this.

jschaedl:
@COil I did, but the alias is not added.

COil:
Weird. I use : 

``` 
populate: ## Reset and populate the elasticsearch index
	$(SYMFONY) fos:elastica:create
	$(SYMFONY) fos:elastica:reset
	$(SYMFONY) fos:elastica:populate --index=app
` ` ` 

You should delete the existing index so it can work.

jschaedl:
@COil I think it is working for you, because you execute `fos:elastica:reset`, which will create the aliases: https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/src/Index/Resetter.php#L119

COil:
I've just verified. And yes the alias is Ok and switched, but the old index is not deleted. I will checl on my other setup.

COil:
You are right, the alias were created with the reset command, not with the create one. I have removed the create command from my setup.

thePanz:
I tested this pull-request on a local instance (SF 4.4, ES 6.8), and it perfectly works! Thank you @jschaedl!
The alias is correctly set when creating a new index.

jschaedl:
🎉

COil:
Thanks @lsmith77  👍

PR #1578, opened 2019-11-28
Add command to delete indexes

jschaedl:


XWB:
It would be nice if you could add a test as well.

COil:
What if it's an alias?

jschaedl:
@COil If it's an alias, all referenced indexes will be deleted.

XWB:
Thanks

PR #1581, opened 2019-12-02
Symfony5 support

tacman:
I've made some changes that allow this bundle to work with Symfony 5.  I'm not familiar with how to handle tests, travis, etc, though.

### DIFF
--- a/src/Command/PopulateCommand.php
+++ b/src/Command/PopulateCommand.php
@@ -185,7 +185,7 @@ class PopulateCommand extends Command
     private function populateIndex(OutputInterface $output, $index, $reset, $options)
     {
         $event = new IndexPopulateEvent($index, $reset, $options);
-        $this->dispatcher->dispatch(IndexPopulateEvent::PRE_INDEX_POPULATE, $event);
+        $this->dispatcher->dispatch($event, IndexPopulateEvent::PRE_INDEX_POPULATE);
 
         if ($event->isReset()) {
             $output->writeln(sprintf('<info>Resetting</info> <comment>%s</comment>', $index));
@@ -197,7 +197,7 @@ class PopulateCommand extends Command
             $this->populateIndexType($output, $index, $type, false, $event->getOptions());
         }
 
-        $this->dispatcher->dispatch(IndexPopulateEvent::POST_INDEX_POPULATE, $event);
+        $this->dispatcher->dispatch($event, IndexPopulateEvent::POST_INDEX_POPULATE);
 
         $this->refreshIndex($output, $index);
     }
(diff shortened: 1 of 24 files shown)


src/Command/PopulateCommand.php
```diff
@@ -185,7 +185,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
     private function populateIndex(OutputInterface $output, $index, $reset, $options)
     {
         $event = new IndexPopulateEvent($index, $reset, $options);
-        $this->dispatcher->dispatch(IndexPopulateEvent::PRE_INDEX_POPULATE, $event);
+        $this->dispatcher->dispatch($event, IndexPopulateEvent::PRE_INDEX_POPULATE);
```
XWB: This won't work on < Symfony 4.3

src/Event/IndexEvent.php
```diff
@@ -11,7 +11,7 @@
 
 namespace FOS\ElasticaBundle\Event;
 
-use Symfony\Component\EventDispatcher\Event;
+use Symfony\Contracts\EventDispatcher\Event;
```
XWB: Older Symfony versions might not have the contracts component.

src/Command/PopulateCommand.php
tacman: So how is this handled then?  A new version of the bundle?  Symfony 4.4 handles it by checking the argument order and calling a legacy handler (and throwing a deprecation warning).  Obviously, we don't want to not support Symfony 5 in favor of supporting older versions of Symfony, so how is this handled? (I don't have much experience with supporting bundles, even branches, tagging and pull requests are new territory for me).

tacman: Um, so what's a better solution for supporting Symfony 5?  A LegacyDispatcher class?  How do you handle Contracts?  Is it something you can do, as this is a key bundle for some of us.  Thanks.

franmomu: I would check if `$dispatcher` implements `Symfony\Contracts\EventDispatcher\EventDispatcherInterface` [as suggested here](https://github.com/liip/LiipImagineBundle/pull/1212#issuecomment-523840038).

src/Event/IndexEvent.php
franmomu: You could create [a `BCEvent` and extend from it](https://github.com/liip/LiipImagineBundle/pull/1218).

franmomu:
I guess this should be done in https://github.com/FriendsOfSymfony/FOSElasticaBundle/tree/5.1.x instead of `master`.

XWB:

PR #1583, opened 2019-12-03
Remove EventDispatcher deprecation warnings

jeroennoten:
- Instead of the deprecated `Symfony\Component\EventDispatcher\Event`, extend the `Symfony\Contract\EventDispatcher\Event` if existent in order to remove deprecation warnings. For Symfony 4.2 and before, we still use `Symfony\Component\EventDispatcher\Event`.

- Add a bundle specific implementation of the `LegacyEventDispatcherProxy` that works for all versions of Symfony and will not throw deprecation warnings.


XWB:

PR #1584, opened 2019-12-03
Use Elasticsearch host and port parameters in all tests

jeroennoten:
Use the Elasticsearch host and port parameters in all tests instead of the default localhost:9200.


XWB:
👍

PR #1573, opened 2019-11-23
allow symfony 5

Jeroeny:
#SymfonyHackday

PR #1570, opened 2019-11-07
fixes deprecation warning of Symfony 4.3.

eroluysal:

PR #1592, opened 2019-12-23
added support for Symfony 5

### LINKED ISSUE #1587: Symfony 5 compatibility check
We can not upgrade to symfony 5 because this bundle does not support it.

Please update the config, if it is compatible with symfony 5.

Thank you!

kuraobi:
This Pull Request enables the bundle to work with Symfony 5. This includes:
- Support for legacy EventDispatcher for Symfony 3.4 as well as new EventDispatcher (contracts, new param order for dispatch) for Symfony 4.3+
- Return values for commands (required, was optional before Symfony 5)
- Support for both legacy and new `DataCollector::collect()` method, which has switched from `Exception` to `Throwable` typehint
- Fix tests for mocked methods with not nullable return types (`UrlGeneratorInterface`, `FragmentHandler`) and class `HttpKernelRuntime` which became final
- Support for new versions of related bundles (`jms/serializer-bundle@^3`, `doctrine/doctrine-bundle@^2`, `knplabs/knp-components@^2`)

Fixes #1587 , also replaces #1583 , #1581 , #1573 , #1570 and #1566 .

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,30 +13,30 @@
     ],
     "require": {
         "php": "^7.1",
-        "symfony/framework-bundle": "^3.4|^4.3",
-        "symfony/console": "^3.4|^4.3",
-        "symfony/dependency-injection": "^3.4|^4.3",
-        "symfony/property-access": "^3.4|^4.3",
+        "symfony/framework-bundle": "^3.4|^4.3|^5",
+        "symfony/console": "^3.4|^4.3|^5",
+        "symfony/dependency-injection": "^3.4|^4.3|^5",
+        "symfony/property-access": "^3.4|^4.3|^5",
         "pagerfanta/pagerfanta": "^1.0.5|^2.0",
         "psr/log": "^1.0",
         "ruflin/elastica": "^5.3.5|^6.1.1"
     },
     "require-dev": {
         "doctrine/orm": "^2.5",
-        "doctrine/doctrine-bundle": "^1.6",
+        "doctrine/doctrine-bundle": "^1.6|^2",
         "doctrine/persistence": "^1.3.3",
         "doctrine/phpcr-bundle": "^1.3|^2.0",
         "doctrine/phpcr-odm": "^1.4",
         "jackalope/jackalope-doctrine-dbal": "^1.2",
-        "jms/serializer-bundle": "^2.2",
-        "phpunit/phpunit": "^5.7.11|^6.5",
-        "knplabs/knp-components": "^1.2",
-        "symfony/expression-language" : "^3.4|^4.3",
-        "symfony/twig-bundle": "^3.4|^4.3",
-        "symfony/serializer": "^3.4|^4.3",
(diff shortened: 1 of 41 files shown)


thePanz:
Is this a duplicate of #1573 ?

kuraobi:
Yes, this PR addresses all the problems related to making the bundle work with Symfony 5. This includes fixing all deprecations, as well as handling both Sf3.4 and Sf5 versions of the EventDispatcher and a few other components, to keep compatibility with all supported Sf versions. I still have a weird error on some configurations but I didn't have the time to look into it.

kuraobi:
I fixed the error, it was actually caused by `jms/serializer` being upgraded, while the functional tests were meant for `jms/serializer@^1`. The test config was broken because the newer versions have switched to snake_case. Everything is green now.

COil:
Hi @lsmith77 , can it be merged please?

composer.json
```diff
@@ -13,30 +13,30 @@
...
         "ruflin/elastica": "^5.3.5|^6.1.1"
     },
     "require-dev": {
         "doctrine/orm": "^2.5",
-        "doctrine/doctrine-bundle": "^1.6",
+        "doctrine/doctrine-bundle": "^1.6|^2",
         "doctrine/persistence": "^1.3.3",
         "doctrine/phpcr-bundle": "^1.3|^2.0",
-        "doctrine/phpcr-odm": "^1.4",
+        "doctrine/phpcr-odm": "^1.5",
```
lsmith77: any reason to no longer allow 1.4 ?

src/DataCollector/ElasticaDataCollector.php
```diff
@@ -15,77 +15,154 @@
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DataCollector\DataCollector;
+use Symfony\Component\HttpKernel\Kernel;
```
lsmith77: maybe it makes more sense to have 2 separate data collector classes one for legacy and one for the future? this way the code is cleaner to read and we can deprecate the legacy class

kuraobi: It would make code cleaner, but it also means we have to change the class name, and ask users to explicitly switch to the new class when migrating to Symfony 5. This approach is not pretty but it does allow for a transparent upgrade from the user. Both solutions have pros and cons so, I don't know.

composer.json
kuraobi: I don't remember, I'll try to revert it to ^1.4 and see if everything works

XWB:
Thank you @kuraobi

amigainc:
Hi, any news about the compatibility with **Symfony 5.***? Some requests are marked as "closed" but still unable to install it using _Composer_.
Regards

erikzogg:
Installation for Symfony 5 was successful with `composer require friendsofsymfony/elastica-bundle:5.x-dev`

PR #1598, opened 2020-01-09
Reverted bc break

COil:
stof : the only BC break in it currently (making it a 6.x-dev) is the findHybridPaginated method in PaginatedFinderInterface. This could be added first as a @method in 5.x (as Symfony does)

### DIFF
--- a/src/Finder/PaginatedFinderInterface.php
+++ b/src/Finder/PaginatedFinderInterface.php
@@ -15,6 +15,9 @@ use Elastica\Query;
 use FOS\ElasticaBundle\Paginator\PaginatorAdapterInterface;
 use Pagerfanta\Pagerfanta;
 
+/**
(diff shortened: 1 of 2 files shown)


XWB:
This should be added to the 5.x branch too. Would you be able to open a new PR @COil?

COil:
But there are several related commit in the master. We have doe this to be able to merge the master in the 5.x branch.

COil:
@XWB   Why don't we merge the master in the 5.x branch ? /cc @lsmith77

XWB:
The master branch is v6, therefore it cannot be merged into v5. The workflow is on the other way around: v5 can be merged into v6.

COil:
But V6 has no meaning. It is V6 if the master contain BC break, if not it is still V5. What are the BC actually in the master ?

PR #1569, opened 2019-11-07
Migrate to ruflin/elastica 7.x

deguif:
- [x] Remove types
- [x] Remove `TypeConfig` class and adapt code related.
- [x] Remove `_doc` type in configuration as it's implicit.
- [x] Move `indexable_callback` configuration to `index` configuration section.
- [x] Move `persistence` configuration to `index` configuration section.
- [x] Move `serializer` configuration to `index` configuration section.
- [x] Remove `_all` meta field configuration (removed in ES 7.0).

PR #1599, opened 2020-01-10
Removed support for Elasticsearch 7

COil:
Elasticsearch 7 and corresponding `ruflin/elastica` library will only be available in next major version.


COil:
Wait, it doesn't work for now.

XWB:
Will be addressed in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1569

PR #1604, opened 2020-02-16
Add finder method to retrieve raw ElasticSearch documents

MichaelDemeyer:
A simple method allowing the ease of use of Finder, and avoiding database calls when entities are not required.

### DIFF
--- a/src/Finder/TransformedFinder.php
+++ b/src/Finder/TransformedFinder.php
@@ -70,6 +70,18 @@ class TransformedFinder implements PaginatedFinderInterface
     }
 
     /**
+     * @param $query
+     * @param null|int $limit
+     * @param array    $options
+     *
+     * @return array
+     */
+    public function findRaw($query, ?int $limit = null, array $options = []): array
+    {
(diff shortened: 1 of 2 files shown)


src/Finder/TransformedFinder.php
```diff
@@ -69,6 +69,18 @@ public function findHybrid($query, $limit = null, $options = [])
...
     }
 
+    /**
+     * @param $query
+     * @param null|int $limit
+     * @param array    $options
+     *
+     * @return array
+     */
+    public function findRaw($query, $limit = null, $option = [])
```
XWB, line 79: You may use type hints and return types.

XWB (changes requested):

src/Finder/TransformedFinder.php
MichaelDemeyer: Fixed.
I do not type hint `$query` because it can be an array or an \Elastica\Query object.

XWB (approved):

XWB:
👍

PR #1613, opened 2020-03-25
resolve symfony deprecations

giovannialbero1992:
This PR introduce:
 - a new Event class that maintain BC to SF3.4
 - dispatch method compliant to new method signature
 - removed by travis dunglas/symfony-lock and added symfony/flex

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,6 @@ env:
 matrix:
   fast_finish: true
   include:
-    # Minimum supported dependencies with min and max PHP version
-    - php: 7.1
-      env: COMPOSER_FLAGS="--prefer-stable --prefer-lowest"
-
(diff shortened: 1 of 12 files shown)


tacman:
PHP 7.1 is no longer supported.  Can we remove it from Travis?  Perhaps this will pass then.

giovannialbero1992:
For me is ok, but it should be clear that we drop compatibility with SF 3.4, the test fail because is installed sf3.4 and dispatch method is different

franmomu:
The [5.x branch already supports Symfony 5](https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/5.x/composer.json), but there is no release yet.

gmorel:
Hi, thanks for your hard work.
Would it be possible to have a tag on 5.x ?

PR #1615, opened 2020-03-31
Remove PHP 7.1 testing

tacman:
Since 7.1 is no longer supported.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,11 +15,10 @@ matrix:
   fast_finish: true
   include:
     # Minimum supported dependencies with min and max PHP version
-    - php: 7.1
+    - php: 7.2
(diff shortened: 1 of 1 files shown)


rushi:
`composer install` won't work because composer.json requires PHP `^7.1` https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/composer.json#L15

Also phpcr/phpcr-utils 1.3.0 requires php ^5.6|7.0.x|7.1.x

deguif:
Hi @tacman 

This should be fixed by https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1569

XWB:
^^

PR #1622, opened 2020-05-02
Support for queued messages via Symfony Messenger

### LINKED ISSUE #1586: Support queued messages via Symfony Messenger?
Now that Symfony Messenger is a first-class component, what do folks think of building in the functionality of EnqueueBundle in this bundle?

It _might_ be something like replacing the ->dispatch calls that are currently events on a configurable message bus.  That would allow us to replace the idea of deferring the indexing altogether, since the user would simply configure how they wanted the dispatched messages handled.

rimas-kudelis:
This kinda fixes #1586, but probably still needs tests and whatnot.

I'm using an ES7-compatible fork of your library at work, and I decided to give this feature a go since we needed to populate a large set of data into ES and are already using Symfony Messenger but not Enqueue. I'm currently using this very persister (with a different namespace) to seed our ES storage.

The way `AsyncPagerPersister` class works (deferrence to `InPlacePagerPersister` to do the actual persister work) was copied from EnqueueElasticaBundle. Basically, I just adapted their algorithm for Messenger and decided to place the `insertPage` method in the same class instead of a separate one.

If you feel like this is useful, feel free to use this PR as a starting ground for your implementation. (In other words, I can't promise I'll be willing to work on it further).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
         "symfony/web-profiler-bundle": "^3.4|^4.0"
     },
     "suggest": {
-        "enqueue/elastica-bundle": "The bundle adds extra features to FOSElasticaBundle bundle. Aimed to improve performance."
+        "symfony/messenger": "For populating Elasticsearch indexes asynchronously and using significanly less resources.",
+        "enqueue/elastica-bundle": "For populating Elasticsearch indexes asynchronously and using significanly less resources. Uses Enqueue."
     },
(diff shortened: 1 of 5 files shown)


XWB:
Would you be able to fix the unit test?

XWB:
Fixed in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1640

PR #1623, opened 2020-05-02
Use Doctrine Paginator, not Pagerfanta, to page data when populating it to ES

rimas-kudelis:
I'd say Pagerfanta is a heavily frontent-oriented library. There's no need to use it when paginating stuff in the backend.


XWB:
Please fix the unit test.

XWB:
Closing as outdated.

PR #1640, opened 2020-09-11
continue - Support for queued messages via Symfony Messenger

giovannialbero1992:
try to continue #1622 @rimas-kudelis

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -40,7 +40,8 @@
         "symfony/web-profiler-bundle": "^4.4 || ^5.1"
     },
     "suggest": {
-        "enqueue/elastica-bundle": "The bundle adds extra features to FOSElasticaBundle bundle. Aimed to improve performance."
+        "symfony/messenger": "For populating Elasticsearch indexes asynchronously and using significanly less resources.",
+        "enqueue/elastica-bundle": "For populating Elasticsearch indexes asynchronously and using significanly less resources. Uses Enqueue."
     },
     "autoload": {
         "psr-4": { "FOS\\ElasticaBundle\\": "src/" }
--- a/composer.json
+++ b/composer.json
@@ -16,6 +16,7 @@
         "symfony/framework-bundle": "^4.4 || ^5.1",
         "symfony/console": "^4.4 || ^5.1",
         "symfony/dependency-injection": "^4.4 || ^5.1",
+        "symfony/messenger": "^4.4 || ^5.1",
         "symfony/property-access": "^4.4 || ^5.1",
         "pagerfanta/pagerfanta": "^2.0",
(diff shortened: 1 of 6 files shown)


XWB:
Looks good! Would it be possible to add a unit test for the new services?

src/Resources/config/persister.xml
```diff
@@ -14,6 +14,14 @@
             <tag name="fos_elastica.pager_persister" persisterName="in_place" />
         </service>
 
+        <service id="fos_elastica.async_pager_persister" class="FOS\ElasticaBundle\Persister\AsyncPagerPersister" public="true">
```
XWB, line 17: Any reason why this service needs to be public?

giovannialbero1992: I've inherited this definition by @rimas-kudelis. It depends if this service could be retrieved by container or it is only injected into services.

rimas-kudelis: @giovannialbero1992 the question was specifically about `public="true"`. You could try removing this attribute. I don't recall why I added it, but it could as well be copy-paste from the Enqueue bundle.

XWB: Yes indeed, I do not see any reason why this service should be public.

giovannialbero1992: @rimas-kudelis Indeed I've answered to XWB about `public="true"` if it is defined, it permits to retrieve service by the container with a `$container->get('fos_elastica.async_pager_persister')` so if it is not necessary I'll remove it in a while

XWB:
Excellent work 👍

composer.json
```diff
@@ -40,7 +41,8 @@
         "symfony/web-profiler-bundle": "^4.4 || ^5.1"
     },
     "suggest": {
-        "enqueue/elastica-bundle": "The bundle adds extra features to FOSElasticaBundle bundle. Aimed to improve performance."
+        "symfony/messenger": "For populating Elasticsearch indexes asynchronously and using significanly less resources.",
```
deguif, line 44: We suggest the package but it's already a direct dependency :thinking:

src/Message/Handler/AsyncPersistPageHandler.php
```diff
@@ -0,0 +1,26 @@
+<?php
+declare(strict_types=1);
```
deguif, line 2: Shouldn't be just after the php opening tag. CS is broken on master.

composer.json
rimas-kudelis: I don't think it was supposed to be a direct dependency.

XWB: Should we remove the dependency?

PR #1641, opened 2020-09-11
Remove types in documentation

deguif:


XWB:
👍

doc/cookbook/manual-provider.md
```diff
@@ -2,15 +2,15 @@ Manual pager provider
 ====================
 
 Create a service with the tag "fos_elastica.pager_provider" and attributes for the
-index and type for which the service will provide.
+index for which the service will provide.
```
kbsali, line 5: Doesn't this sound a bit strange?

doc/indexes.md
```diff
@@ -1,19 +1,21 @@
...
+Index configuration
 ==================
 
 Custom Property Paths
 ---------------------
 
-Since FOSElasticaBundle 3.1.0, it is now possible to define custom property paths
-to be used for data retrieval from the underlying model.
+Custom property paths can be used for data retrieval from the underlying model.
+Custom property paths can be used for data retrieval from the underlying model.
```
kbsali, line 8: 1 time only! ;)

doc/setup.md
```diff
@@ -79,75 +79,54 @@ In this case, the service `fos_elastica.index.app` will relate to an Elasticsear
...
-                        title: ~
-                        content: ~                       
-```
+Each defined index is made available as a service, and in this case the service key is
+`fos_elastica.index.user` and is an instance of `Elastica\Index`.
 
-FOSElasticaBundle requires a provider for each type that will notify when an object
-that maps to a type has been modified. The bundle ships with support for Doctrine objects.
+FOSElasticaBundle requires a provider for each index that will notify when an object
+that maps to an index has been modified. The bundle ships with support for Doctrine objects.
```
kbsali, line 107: `that maps an index`

kbsali:
:+1:

PR #1643, opened 2020-09-14
Tweak documentation

deguif:
Apply comments from #1641

### DIFF
--- a/doc/cookbook/custom-properties.md
+++ b/doc/cookbook/custom-properties.md
@@ -5,7 +5,7 @@ object into an Elastica document which allows you to set custom properties on th
 document for indexing.
 
 Set up an event listener or subscriber for 
-`FOS\ElasticaBundle\Event\TransformEvent::POST_TRANSFORM` to be able to inject your own
+`FOS\ElasticaBundle\Event\PostTransformEvent` to be able to inject your own
 parameters.
 
 ```php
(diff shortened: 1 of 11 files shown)


doc/cookbook/custom-properties.md
```diff
@@ -31,14 +31,14 @@ class CustomPropertyListener implements EventSubscriberInterface
     
     public static function getSubscribedEvents()
     {
-        return array(
-            TransformEvent::POST_TRANSFORM => 'addCustomProperty',
-        );
+        return [
+            PostTransformEvent::class => 'addCustomProperty',
```
XWB, line 35: I haven't seen this before, what makes this different from `TransformEvent::POST_TRANSFORM`?

deguif: That's the new standard for Symfony event dispatcher if I'm not wrong.
Use class name as event name, when it's possible (ie. not same class used for different events).

deguif: Just found an article that explained the move: https://symfony.com/blog/new-in-symfony-4-3-simpler-event-dispatching

deguif: By the way in the previous PR adding support for elastica, I removed all these constants and used different classes for each event, so that should now be the only way to listen / subscribe on these events.

XWB: Thank you.

XWB: > By the way in the previous PR adding support for elastica, I removed all these constants and used different classes for each event, so that should now be the only way to listen / subscribe on these events.

We should also list those changes in https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/CHANGELOG-6.0.md

deguif: Yes I see that there is a line about the `FOS\ElasticaBundle\Persister\Event\Events` but more details should probably be added to facilitate the move for other events

deguif: Done in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1657

PR #1644, opened 2020-09-14
Use short array syntax everywhere

deguif:


XWB:
:+1:

deguif:
We should probably add a CS rule to ensure that ;)

XWB:
We already have that rule: https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/.php_cs.dist#L15

PR #1653, opened 2020-09-14
Add type-hints on signatures of client and logger methods

deguif:


src/Elastica/Client.php
```diff
@@ -51,7 +51,7 @@ class Client extends BaseClient
     /**
      * {@inheritdoc}
      */
-    public function request($path, $method = Request::GET, $data = [], array $query = [], $contentType = Request::DEFAULT_CONTENT_TYPE): Response
+    public function request(string $path, string $method = Request::GET, $data = [], array $query = [], string $contentType = Request::DEFAULT_CONTENT_TYPE): Response
```
deguif, line 54: This was changed in `ruflin/elastica`.

PR #1657, opened 2020-09-14
Update changelog

deguif:


deguif:
@XWB I think we're ready to publish a beta version.

Here's a remaining part I would like to think before the stable version.
Currently there are two ways to configure a finder for an index in the configuration (the second one is kept back from type configuration but can probably be dropped or re-adapted with the first one).

XWB:
> Here's a remaining part I would like to think before the stable version.

I think we should also get rid of all public services.

CHANGELOG-6.0.md
```diff
@@ -11,19 +11,35 @@ in 6.0 versions.
...
+* **[BC break]** Changed signature of methods `findPaginated`, `createPaginatorAdapter` and `createRawPaginatorAdapter` from interface `FOS\ElasticaBundle\Finder\PaginatedFinderInterface`.
+* **[BC break]** Changed signature of methods `request`, `getIndex`, `getIndexTemplate` and `setStopwatch` from class `FOS\ElasticaBundle\Elastica\Client`.
+* **[BC break]** Changed signature of methods `logQuery`, `getNbQueries` and `getQueries` from class `FOS\ElasticaBundle\Logger\ElasticaLogger`.
+* **[BC break]** Changed signature of methods `addIndex`, `getRepository` and `getRepositoryName` from class `FOS\ElasticaBundle\Manager\RepositoryManager`.
+* **[BC break]** Changed signature of methods `find`, `findHybrid`, `findPaginated` and `createPaginatorAdapter` from class `FOS\ElasticaBundle\Repository`.
+* **[BC break]** Removed `Elastica\Type`.
+* **[BC break]** Removed `_parent`.
+* **[BC Break]** Removed `FOS\ElasticaBundle\Event\IndexPopulateEvent` constants for event names, use event classes instead:
```
XWB, line 28: I think "Break" should be lowercase.

deguif: Yes my bad ;)

deguif: Fixed

PR #1662, opened 2020-09-15
Mark repository managers private

XWB:


tests/Functional/PersistenceRepositoryTest.php
```diff
@@ -15,9 +15,15 @@ class PersistenceRepositoryTest extends WebTestCase
 {
     public function testRepositoryShouldBeSetCorrectly()
     {
-        static::bootKernel(['test_case' => 'ORM']);
+        selfi::bootKernel(['test_case' => 'ORM']);
```
deguif, line 18: selfi ;)

XWB: rofl :)

CHANGELOG-6.0.md
```diff
@@ -43,3 +43,4 @@ in 6.0 versions.
     - `POST_PERSIST` => `FOS\ElasticaBundle\Persister\Event\PostPersistEvent`
     - `PRE_FETCH_OBJECTS` => `FOS\ElasticaBundle\Persister\Event\PreFetchObjectsEvent`
 * **[BC break]** Renamed `FOS\ElasticaBundle\Persister\Event\OnExceptionEvent::setIgnore()` to `FOS\ElasticaBundle\Persister\Event\OnExceptionEvent::setIgnored()`.
+* **[BC break]** All `fos_elastica.manager` services have been marked private.
```
deguif, line 46:
```suggestion
* **[BC break]** Marked all `fos_elastica.manager` services as private.
```

deguif:
@XWB can you try setting `test: true` under `framework` inside `tests/Functional/app/config/config.yml`

deguif:
You can do it as well in `tests/Functional/app/ORM/config.yml`

tests/Functional/PersistenceRepositoryTest.php
```diff
@@ -15,9 +15,15 @@ class PersistenceRepositoryTest extends WebTestCase
...
     {
-        static::bootKernel(['test_case' => 'ORM']);
+        self::bootKernel(['test_case' => 'ORM']);
 
-        $repository = static::$kernel->getContainer()->get('fos_elastica.manager.orm')
+        // returns the real and unchanged service container
+        $container = self::$kernel->getContainer();
+
+        // gets the special container that allows fetching private services
+        $container = self::$container;
```
deguif, line 24: Can we add a debug in the test to see what's the class of this object?

XWB: Yes something weird is going on. I'm doing just the same as the cookbook: https://symfony.com/blog/new-in-symfony-4-1-simpler-service-testing yet the test keeps failing.

XWB: I'm getting `Symfony\Bundle\FrameworkBundle\Test\TestContainer`

deguif:
```
Keep in mind that, because of how Symfony's service container work, unused services are removed from the container. This means that if you have a private service not used by any other service, Symfony will remove it and you won't be able to get it as explained in this article. The solution is to define the service as public explicitly so Symfony doesn't remove it.
```

XWB: But I'm using the service in the test :P

tests/Functional/app/ORM/config.yml
```diff
@@ -10,6 +10,8 @@ doctrine:
         auto_mapping: false
 
 services:
+    _defaults:
+        public: true
```
deguif, line 14: That should only apply to service declared here, that's why it doesn't work ;)

deguif: Can you add 
```yml
fos_elastica.manager.orm:
    public: true
```

tests/Functional/PersistenceRepositoryTest.php
deguif: Yes but the container is already compiled when trying to retrieve the service.

tests/Functional/app/ORM/config.yml
deguif: I'm pretty sure this will fix the error ;)

XWB: Perhaps this might work as well:

```yaml
services:
    _defaults:
        public: true
    test_alias.fos_elastica.indexable:
        alias: fos_elastica.indexable
    indexable_service:
        class: FOS\ElasticaBundle\Tests\Functional\app\ORM\IndexableService
    fos_elastica.manager.orm:
        class: FOS\ElasticaBundle\Doctrine\RepositoryManager
```

deguif: Yes because these service declared here will be public, so this should work as well ;)

tests/Functional/app/ORM/config.yml
```diff
@@ -16,8 +16,8 @@ services:
         alias: fos_elastica.indexable
     indexable_service:
         class: FOS\ElasticaBundle\Tests\Functional\app\ORM\IndexableService
-    fos_elastica.manager.orm:
-        class: FOS\ElasticaBundle\Doctrine\RepositoryManager
+    test_alias.fos_elastica.manager.orm:
+        alias: fos_elastica.manager.orm
```
deguif, line 20: Should better work now ;)

XWB: Yes tests are finally passing :P

deguif: Maybe we can set only alias to public true and avoid setting public the indexable_service?

XWB: Like this: https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1664 ?

deguif: Yes but sure the indexable_service can be private ;)

XWB: Well, it was private before so it should still work.

PR #1663, opened 2020-09-15
Improve exception messages

deguif:


src/Command/DeleteCommand.php
```diff
@@ -78,7 +78,7 @@ private function deleteIndex(string $indexName): void
             $path = sprintf('%s', $indexName);
             $this->client->request($path, Request::DELETE);
         } catch (ExceptionInterface $deleteOldIndexException) {
-            throw new \RuntimeException(sprintf('Failed to delete index %s with message: %s', $indexName, $deleteOldIndexException->getMessage()), 0, $deleteOldIndexException);
+            throw new \RuntimeException(sprintf('Failed to delete index "%s" with message: "%s"', $indexName, $deleteOldIndexException->getMessage()), 0, $deleteOldIndexException);
```
deguif, line 81: I would be in favor of adding a `.` at the end as it's also used sometimes but not always ;)

PR #1665, opened 2020-09-15
Remove superflous sprintf

deguif:


src/Command/DeleteCommand.php
```diff
@@ -75,7 +75,7 @@ protected function execute(InputInterface $input, OutputInterface $output)
     private function deleteIndex(string $indexName): void
     {
         try {
-            $path = sprintf('%s', $indexName);
+            $path = $indexName;
```
deguif, line 78: The variable can be removed too, but I think it was here to more understandable?

XWB: > it was here to be more understandable

Yes indeed.

PR #1667, opened 2020-09-15
Some more cleanup

XWB:


src/Manager/RepositoryManagerInterface.php
```diff
@@ -26,8 +26,7 @@ interface RepositoryManagerInterface
      * Adds index name and its finder.
      * Custom repository class name can also be added.
      *
-     * @param string $indexName      The type name in "index/type" format
-     * @param string $repositoryName
+     * @param string $indexName The type name in "index/type" format
```
deguif, line 29:
```suggestion
     * @param string $indexName The type name in "index" format
```

But useless I think

deguif (approved):

src/Manager/RepositoryManagerInterface.php
```diff
@@ -25,9 +25,6 @@ interface RepositoryManagerInterface
     /**
      * Adds index name and its finder.
      * Custom repository class name can also be added.
-     *
-     * @param string $indexName      The type name in "index/type" format
-     * @param string $repositoryName
      */
     public function addIndex(string $indexName, FinderInterface $finder, string $repositoryName = null): void;
```
deguif, line 29:
```suggestion
    public function addIndex(string $indexName, FinderInterface $finder, ?string $repositoryName = null): void;
```

deguif: There is a CS rule for that: `nullable_type_declaration_for_default_null_value`

deguif: I remember adding it to ruflin/elastica ;)

XWB: Nice. PR has been updated :)

PR #1676, opened 2020-09-16
Run tests on both symfony versions supported

giovannialbero1992:
Run tests on Symfony 4.4 avoiding to use a bad version (like symfony/stopwatch 3.4 with prefer-lowest option)


XWB:
Good catch :+1:

PR #1678, opened 2020-09-16
Add new cs check for composer.json

deguif:


.github/workflows/continuous-integration.yaml
```diff
@@ -12,10 +12,19 @@ jobs:
         uses: 'shivammathur/setup-php@v2'
         with:
           php-version: '7.4'
+          coverage: 'none'
           extensions: 'json, mbstring, tokenizer'
-          tools: 'cs2pr, php-cs-fixer:2.16.4'
```
deguif, line 16: Removing `cs2pr` as it doesn't bring much value with php-cs-fixer as it doesn't indicate line numbers

PR #1683, opened 2020-09-19
options added to findHybridPaginated

geoffroyp:


deguif:
See https://github.com/FriendsOfSymfony/FOSElasticaBundle/runs/1137480470?check_suite_focus=true for failures 
This is related to coding styles

geoffroyp:
Yeah I figured that it was the list of stuff that were "wrong", but I don't get what is wrong. Is there a doc about the coding style you are using that may give me a hint, or something?

deguif:
You have the list of failures in the step `Chech PHP code` on the link I gave you previously. The coding styles are in the `.php_cs.dist` file at the root of the project.
But you have the diff for fixing.

XWB:
The solution is simple: simple run cs-fixer and commit the changes.

geoffroyp:
ok so apparently it was all the phpdoc about this new "options" argument I added that caused the coding style problem.

I'm genuinely curious why the coding style you've set accept some argument (like the query one), but reject other (like the options array). If someone has time and is kind enough to tell me, I'd be happy!

src/Paginator/HybridPaginatorAdapter.php
```diff
@@ -27,9 +27,9 @@ class HybridPaginatorAdapter extends RawPaginatorAdapter
      * @param Query                               $query       the query to search
      * @param ElasticaToModelTransformerInterface $transformer the transformer for fetching the results
      */
-    public function __construct(SearchableInterface $searchable, Query $query, ElasticaToModelTransformerInterface $transformer)
+    public function __construct(SearchableInterface $searchable, Query $query, array $options, ElasticaToModelTransformerInterface $transformer)
```
XWB, line 30: Wouldn't it be better to pass $options as 3rd argument?

geoffroyp: I tried to stick to the similar functions signature that were in the same file. But yes,I guess that could cause BC break. But on the other hand it makes this function consistent with the others.

You tell me

geoffroyp: I guess. I'll change it now

deguif:
`no_superfluous_phpdoc_tags` that's rhe rule you were asking for.
See https://mlocati.github.io/php-cs-fixer-configurator/#version:2.16|fixer:no_superfluous_phpdoc_tags

geoffroyp:
got it! Thanks @deguif

src/Paginator/HybridPaginatorAdapter.php
XWB: BC breaks are allowed in master branch as long as 6.0 is not final.

PR #1687, opened 2020-09-21
Make in_place_pager_persister public

### LINKED ISSUE #1686: Service or alias has been removed in populate
In dev-master version bin/console fos:elastica:populate throws error:

In Container.php line 266:
                                                                                                                                                                                                                                 
  The "fos_elastica.in_place_pager_persister" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.

tacman:
Fixes #1686

### DIFF
--- a/src/Resources/config/persister.xml
+++ b/src/Resources/config/persister.xml
@@ -5,7 +5,7 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
-        <service id="fos_elastica.in_place_pager_persister" class="FOS\ElasticaBundle\Persister\InPlacePagerPersister">
+        <service id="fos_elastica.in_place_pager_persister" class="FOS\ElasticaBundle\Persister\InPlacePagerPersister" public="true">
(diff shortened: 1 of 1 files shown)


XWB:
Symfony does not recommend public services anymore. You should inject the service instead. All public services will be removed in v6 of this bundle.

tacman:
Agreed, but I wanted to propose something to fix the immediate problem, since the populate command throws an error right now.

It looks like the problem is that PagerPersisterRegistration uses ContainerAwareTrait and gets the services by id.  I'm not sure how to get the service without the container in this situation.

XWB:
I think we need to modify `RegisterPagerPersistersPass` to get it fixed.

deguif:
I'm going to propose another approach that uses a `ServiceLocator` instead, that should fix all the problems, and remove the need for `RegisterPagerPersistersPass`

XWB:
Ah yes, the ServiceLocator. Good suggestion, forget about its existence :+1:

XWB:
Should be fixed in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1689

PR #1702, opened 2020-09-28
Fix messenger integration

### LINKED ISSUE #1701: The service "fos_elastica.async_pager_persister" has a dependency on a non-     existent service "messenger.default_bus"
Since upgrade to BETA 2, I got this error.
That's related to new integration with symfony/messenger, but not using symfony/messenger shouldn't break.

I think something is wrong with current implementation. ping @giovannialbero1992

deguif:
Fix #1701


deguif:
@XWB Can we publish another beta version with this fix?

XWB:
@deguif Was already working on it :)

deguif:
Nice, thank you

Issue #1637, opened 2020-09-10
Make the finder services public, so they can be injected and TypeHinting


tacman:
The documentation on how to search assumes the container is available, which is not longer very common in Symfony 5 apps.

    $finder = $this->container->get('fos_elastica.finder.app.user');

https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/doc/usage.md

If these were services were public, they could be injected via typehinting and variable naming.

Alternatively (or additionally), maybe there's a way to get the finder service from a service that is already public.  If so, what is that?  I'll update the documentation once I learn how to do it.  Thanks.

XWB:
You don't have to make the service public. Just inject it via service configuration:

```yaml
    # services.yaml
services:

    App\Elastic\MyService:
        arguments:
            $finder: '@fos_elastica.finder.app.user'
```

```php
namespace App\Elastic;

use FOS\ElasticaBundle\Finder\PaginatedFinderInterface;

class MyService
{
    private PaginatedFinderInterface $finder;

    public function __construct(PaginatedFinderInterface $finder)
    {
        $this->finder = $finder;
    }
}
```

tacman:
Slightly less setup is to bind the parameters:
```yaml
services:
    _defaults:
        autowire: true      # Automatically injects dependencies in your services.
        autoconfigure: true # Automatically registers your services as commands, event subscribers, etc.
        bind:
            $finderWork: '@fos_elastica.finder.work'
            $finderParagraph: '@fos_elastica.finder.paragraph'
```

Then in the service or controller, use that variable

```php
    /**
     * @Route("/search", name="search_dashboard")
     */
    public function search(Request $request, FinderInterface $finderParagraph, FinderInterface $finderWork, IndexManager $indexManager)
    {
```

BUT if they were public, these services could be used without configuring them in services.yaml, by using the camelcase version of the service name __construct(FinderInterface $fosElasticaFinderWork) I think.  This is how monolog channels work, for example.  

I think they should be public, though, because that's how they're intended to be used, and it greatly simplifies the documentation and DX.

tacman:
Ah, getting closer.  Inject the RepositoryManager instead, which is already public.

That returns a Repository object, which has a protected $finder (a PaginatedFinder), and many of the methods.

Since the documentation shows how to get results from the $finder, not a Repository object, it seems like the finder object should be public.

Should I update the documentation and submit a PR?

deguif:
[...]

PR #1706, opened 2020-09-28
Remove remaining public services

XWB:
@deguif Should we also make a service locator for the finder services?

### DIFF
--- a/src/DependencyInjection/FOSElasticaExtension.php
+++ b/src/DependencyInjection/FOSElasticaExtension.php
@@ -15,7 +15,6 @@ use Elastica\Client as ElasticaClient;
 use FOS\ElasticaBundle\Elastica\Client;
 use FOS\ElasticaBundle\Manager\RepositoryManagerInterface;
 use Symfony\Component\Config\FileLocator;
-use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\ContainerAwareInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
@@ -93,14 +92,16 @@ class FOSElasticaExtension extends Extension
         }
 
         $this->loadClients($config['clients'], $container);
-        $container->setAlias('fos_elastica.client', \sprintf('fos_elastica.client.%s', $config['default_client']));
-        $container->getAlias('fos_elastica.client')->setPublic(true);
-        $container->setAlias(ElasticaClient::class, new Alias('fos_elastica.client', false));
-        $container->setAlias(Client::class, 'fos_elastica.client');
-        $container->getAlias(Client::class)->setPublic(false);
+        $container->setAlias('fos_elastica.client', \sprintf('fos_elastica.client.%s', $config['default_client']))
+            ->setPublic(false);
+        $container->setAlias(ElasticaClient::class, 'fos_elastica.client')
+            ->setPublic(false);
+        $container->setAlias(Client::class, 'fos_elastica.client')
+            ->setPublic(false);
 
         $this->loadIndexes($config['indexes'], $container);
-        $container->setAlias('fos_elastica.index', \sprintf('fos_elastica.index.%s', $config['default_index']));
+        $container->setAlias('fos_elastica.index', \sprintf('fos_elastica.index.%s', $config['default_index']))
+            ->setPublic(false);
         $container->setParameter('fos_elastica.default_index', $config['default_index']);
 
         if ($usedIndexNames = \array_intersect_key($config['indexes'], $config['index_templates'])) {
@@ -132,10 +133,8 @@ class FOSElasticaExtension extends Extension
      *
      * @param array            $clients   An array of clients configurations
      * @param ContainerBuilder $container A ContainerBuilder instance
-     *
-     * @return array
      */
-    private function loadClients(array $clients, ContainerBuilder $container)
+    private function loadClients(array $clients, ContainerBuilder $container): void
     {
         foreach ($clients as $name => $clientConfig) {
             $clientId = \sprintf('fos_elastica.client.%s', $name);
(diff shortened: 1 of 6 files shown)


src/DependencyInjection/FOSElasticaExtension.php
```diff
@@ -94,10 +94,10 @@ public function load(array $configs, ContainerBuilder $container)
 
         $this->loadClients($config['clients'], $container);
         $container->setAlias('fos_elastica.client', \sprintf('fos_elastica.client.%s', $config['default_client']));
-        $container->getAlias('fos_elastica.client')->setPublic(true);
+        $container->getAlias('fos_elastica.client')->setPrivate(true);
```
deguif, line 97: `setPrivate()` should not be used since Symfony 3.4, use `setPublic(false)`
See https://github.com/symfony/symfony/pull/37373

src/DependencyInjection/FOSElasticaExtension.php
```diff
@@ -94,10 +94,10 @@ public function load(array $configs, ContainerBuilder $container)
...
         $container->setAlias(ElasticaClient::class, new Alias('fos_elastica.client', false));
         $container->setAlias(Client::class, 'fos_elastica.client');
-        $container->getAlias(Client::class)->setPublic(false);
+        $container->getAlias(Client::class)->setPrivate(true);
```
deguif, line 100: Same here

src/DependencyInjection/FOSElasticaExtension.php
```diff
@@ -564,7 +565,7 @@ private function loadIndexListener(array $indexConfig, ContainerBuilder $contain
         }
 
         if ($indexConfig['listener']['defer']) {
-            $listenerDef->setPublic(true);
+            $listenerDef->setPrivate(true);
```
deguif, line 568: Same here

deguif:
@XWB why would a service locator be needed?

XWB:
@deguif Making the finder services private means you won't be able to fetch it from the container:

`$this->container->get('fos_elastica.finder.app.user');`

The alternative solution was posted in https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1637#issuecomment-690734478 but perhaps a service locator would simplify fetching the finder servives?

deguif:
@XWB the alternative posted in https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1637#issuecomment-690734478 seems the right solution. Service locator wouldn't simplify the retrieval for this kind of thing.

src/DependencyInjection/FOSElasticaExtension.php
```diff
@@ -94,10 +94,8 @@ public function load(array $configs, ContainerBuilder $container)
 
         $this->loadClients($config['clients'], $container);
         $container->setAlias('fos_elastica.client', \sprintf('fos_elastica.client.%s', $config['default_client']));
-        $container->getAlias('fos_elastica.client')->setPublic(true);
-        $container->setAlias(ElasticaClient::class, new Alias('fos_elastica.client', false));
+        $container->setAlias(ElasticaClient::class, 'fos_elastica.client');
```
deguif, line 97: It is still needed to mark it as public false I think for the moment ;)

XWB: Hmm, you're right. Even in 5.1 the default value is public: true https://github.com/symfony/dependency-injection/blob/5.1/Alias.php#L25

I was under the impression all services were private by default since Symfony 3.4: https://symfony.com/blog/new-in-symfony-3-4-services-are-private-by-default

deguif: services yes but not aliases ;)

XWB: Well, the Symfony blog says:

> That's why in Symfony 3.4 services **and aliases** will be private by default.

deguif: You're right but that's for aliases declared in configuration files not by using the container builder ;)

XWB: I see.

Well that was confusing :) 

BTW is there any reason to use the Alias class?

```php
$container->setAlias(ElasticaClient::class, new Alias('fos_elastica.client', false));
```

vs

```php
$container->setAlias(ElasticaClient::class, 'fos_elastica.client')
    ->setPublic(false);
```

deguif:
```php
$container->setAlias(ElasticaClient::class, new Alias('fos_elastica.client', false));
```

That's what some bundle use, so that's fine and less confusing in this case ;)

XWB:
@deguif PR has been updated.

src/DependencyInjection/FOSElasticaExtension.php
```diff
@@ -93,14 +92,16 @@ public function load(array $configs, ContainerBuilder $container)
...
 
         $this->loadClients($config['clients'], $container);
-        $container->setAlias('fos_elastica.client', \sprintf('fos_elastica.client.%s', $config['default_client']));
-        $container->getAlias('fos_elastica.client')->setPublic(true);
-        $container->setAlias(ElasticaClient::class, new Alias('fos_elastica.client', false));
-        $container->setAlias(Client::class, 'fos_elastica.client');
-        $container->getAlias(Client::class)->setPublic(false);
+        $container->setAlias('fos_elastica.client', \sprintf('fos_elastica.client.%s', $config['default_client']))
+            ->setPublic(false);
+        $container->setAlias(ElasticaClient::class, 'fos_elastica.client')
```
deguif, line 97: You don't choose `$container->setAlias(ElasticaClient::class, new Alias('fos_elastica.client', false))` finally ?

XWB: Well, the other aliases https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/src/DependencyInjection/FOSElasticaExtension.php#L99-#L100 do not use the Alias class either, so it was more easier and more consistent to remove this one.

deguif: You're right, can be improved in another PR to make this consitstent.

PR #1707, opened 2020-09-28
Allow PHPUnit 9.3 in order to prepare support for PHP 8.0

deguif:


deguif:
There are some deprecations to fix.
I will update this PR once they are fixed.

deguif:
@XWB all deprecations introduced in PHPUnit 9.x should now be fixed.
I introduced a kind of hack to support both versions of 8.5 and 9.3 regarding `prophesize` deprecation with the `ProphecyTrait`

XWB:
:+1:

PR #1719, opened 2020-10-08
auth_type confguration

thomaspicquet:
its not released yet but this PR will cover the new param when its ready at Elastica https://github.com/ruflin/Elastica/pull/1790


XWB:
:+1:

PR #1727, opened 2020-11-08
updating docs, adding hint for adding multiple queries to bool

leberknecht:
Adding a small note about different behavior of ES 7.6 and 7.7 , closes #1727


XWB:
Thank you.

PR #1735, opened 2020-11-30
Allow Symfony 5.2

darthf1:
Hi! Can we allow SF 5.2?

https://github.com/symfony/symfony/releases/tag/v5.2.0


XWB:
It should already work, this change is not required.

darthf1:
You're correct, it already does.

PR #1736, opened 2020-11-30
Explicitly mark public services as public for Symfony 5.2 compatibility

bobvandevijver:
Symfony 5.2 removed the `setPrivate(false)` call from the `ChildDefinition` class (https://github.com/symfony/dependency-injection/commit/93123c0f564828ee0df027fbf98ad0b084f0966d#diff-053e10444fda1f5cfb6c16eee77d265e49962ec674d505e036a7238b78af8eb4), causing this bundle to throw exception on own registered services during its compiler passes. 

I know that with version 6 of this bundle most services are marked as private, but for the people not wanting to upgrade to v6 (yet), this can be a small service release of the 5.2 version.


XWB:
Seems fair to me.

PR #1738, opened 2020-12-01
Review MappingBuilder

deguif:


src/Index/MappingBuilder.php
```diff
@@ -93,28 +87,19 @@ public function buildMapping(?string $model, IndexConfigInterface $indexConfig)
             $mapping['_meta']['model'] = $model;
         }
 
-        if (empty($mapping)) {
-            // Empty mapping, we want it encoded as a {} instead of a []
-            $mapping = new \ArrayObject();
```
deguif, line 98: Never used in the code base as there is a check before to see if array is empty.
This allows to always have a single return type, `array` now. 

If needed when using this method, the return can be encapsulated in an `ArrayObject` and would provide the same feature.

For example:
```php
$mapping = new \ArrayObject($builder->buildMapping());
```

XWB: Fine by me :+1:

PR #1740, opened 2020-12-01
Make PostMappingBuildEvent extends IndexEvent

deguif:
This would allow to easily check if event needs to be handled for a specific index, by using for example:

```php
if ('my_index' === $event->getIndex()) {
    // do stuff
}
```

### DIFF
--- a/src/Event/PostMappingBuildEvent.php
+++ b/src/Event/PostMappingBuildEvent.php
@@ -12,9 +12,8 @@
 namespace FOS\ElasticaBundle\Event;
 
 use FOS\ElasticaBundle\Configuration\IndexConfigInterface;
-use Symfony\Contracts\EventDispatcher\Event;
 
-final class PostMappingBuildEvent extends Event
+final class PostMappingBuildEvent extends IndexEvent
 {
     /**
      * @var IndexConfigInterface
@@ -30,6 +29,8 @@ final class PostMappingBuildEvent extends Event
     {
         $this->indexConfig = $indexConfig;
         $this->mapping = $mapping;
+
+        parent::__construct($indexConfig->getName());
     }
 
     public function getIndexConfig(): IndexConfigInterface
--- a/src/Event/PostIndexMappingBuildEvent.php
+++ b/src/Event/PostIndexMappingBuildEvent.php
@@ -13,7 +13,7 @@ namespace FOS\ElasticaBundle\Event;
 
 use FOS\ElasticaBundle\Configuration\IndexConfigInterface;
 
-final class PostMappingBuildEvent extends IndexEvent
+final class PostIndexMappingBuildEvent extends IndexEvent
 {
     /**
      * @var IndexConfigInterface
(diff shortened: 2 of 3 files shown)


src/Event/PostMappingBuildEvent.php
```diff
@@ -12,9 +12,8 @@
 namespace FOS\ElasticaBundle\Event;
 
 use FOS\ElasticaBundle\Configuration\IndexConfigInterface;
-use Symfony\Contracts\EventDispatcher\Event;
 
-final class PostMappingBuildEvent extends Event
+final class PostMappingBuildEvent extends IndexEvent
```
deguif, line 16: Maybe we should also rename it to PostIndexMappingBuildEvent?

ping @darthf1

src/Event/PostMappingBuildEvent.php
```diff
@@ -30,6 +29,8 @@ public function __construct(IndexConfigInterface $indexConfig, array $mapping)
     {
         $this->indexConfig = $indexConfig;
         $this->mapping = $mapping;
+
+        parent::__construct($indexConfig->getName());
```
deguif, line 33: I'm always confused about which name to use, should it be `$indexConfig->getElasticsearchName()`?
ping @XWB

src/Event/PostMappingBuildEvent.php
darthf1: I have no opinion on this. It makes sense if you extend the `IndexEvent`

XWB: Yeah, `PostIndexMappingBuildEvent` would be better.

src/Event/PostMappingBuildEvent.php
XWB: There is no `getElasticsearchName` method, do you mean renaming `getName()` to `getElasticsearchName()`?

deguif: I mean this method: https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/src/Configuration/IndexConfigInterface.php#L21 vs this one https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/src/Configuration/IndexConfigInterface.php#L25

There is an abstraction of the name of the real index name that can be achieved with the bundle configuration, when using for example:
```yaml
fos_elastica:
    indexes:
        app:
            index_name: app_prod
```

`getName()` would return in this case `app` where `getElasticsearchName()` would return `app_prod`

XWB: Good question. I have to admin I'm confused too. Perhaps we should stick with `getName()` as long as things work nicely.

deguif: Yes, seems fair ;)

deguif:
Oups it was merged too quickly, I had to rebase before to take into account other PR. I'm creating a new one.

PR #1734, opened 2020-11-29
Add post mapping builder event

darthf1:
Fixes https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1725

PR #1741, opened 2020-12-01
Make event classes final

darthf1:
As mentioned in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1734#discussion_r532887585

### DIFF
--- a/src/Event/PostIndexPopulateEvent.php
+++ b/src/Event/PostIndexPopulateEvent.php
@@ -11,6 +11,6 @@
 
 namespace FOS\ElasticaBundle\Event;
 
-class PostIndexPopulateEvent extends IndexPopulateEvent
+final class PostIndexPopulateEvent extends IndexPopulateEvent
 {
 }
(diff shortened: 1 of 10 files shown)


src/Event/PostIndexPopulateEvent.php
```diff
@@ -11,6 +11,6 @@
 
 namespace FOS\ElasticaBundle\Event;
 
-class PostIndexPopulateEvent extends IndexPopulateEvent
+final class PostIndexPopulateEvent extends IndexPopulateEvent
```
deguif, line 14: Maybe we should also mark the corresponding base class as `abstract`?

darthf1: done

darthf1: Hm tests fail if I add abstract. Should the tests be changed or should we refrain from adding abstract? 

https://travis-ci.com/github/FriendsOfSymfony/FOSElasticaBundle/jobs/451397287

XWB: An abstract class should not be initialized.

If the class was not meant to be abstract, removing the keyword should be enough.

darthf1: Ok, then its done. The result is no abstract classes.

PR #1742, opened 2020-12-01
Mark base event classes as abstract

deguif:


XWB:
A much better solution :+1:

PR #1743, opened 2020-12-07
Fix wrong merge

deguif:


deguif:
It seems the project has some issue with travis now, as tests were not run.

Issue #1732, opened 2020-11-25
Provide parameters in `indexable_callback`


darthf1:
Hi! I'm not sure if this is a feasible request, its sort of a continuation of https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1717

All my entities are related to a `Project` entity. When a user does a certain action for `Project.id = 1`, I want to re-index only objects related to `Project.id = 1`.

The current `indexable_callback` is already powerful; I can call whatever class/function I want. What I don't understand is how I can make the `indexable_callback` function aware that `Project.id` should be `1`. The only parameter currently provided to the callback method is the `$object` being indexed.

An option could be to just write the `Project.id` to a file, and check that file in the `indexable_callback` function. But I cant imagine this being a final solution. I thought about defining a project parameter in the container just before calling the populate command, and have that parameter injected into the callback class. But parameters need to be set before the container is compiled.

What I can imagine, that it would be possible to provide an `options` parameter to the `fos:elastica:populate` command, and whatever you put in there gets passed as a parameter into the `indexable_callback` function next to the object being indexed. I'm not sure if this could be made possible.

What would be the best way to tackle this?

PR #1757, opened 2021-02-11
feat: allow to pass additional options to the provider

darthf1:
This resolves https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1732.

This PR adds an additional parameter to the populate command called `options`. My use case is that I almost never want to fully re-index, but just want to add new documents to the index. The additional options provided, like `--options=project=1`, are passed to the provider. With a custom provider, I can make sure to only Index documents where project=1 (when `--no-reset` is set as well).

These options are converted from `--options=key=value` to `["key" => "value"]`, merged with the other options, and then passed through to the provider.


darthf1:
So what are your thoughts about this PR?

PR #1761, opened 2021-03-12
Allow to disable aliasing when creating index

deguif:
This is complementary to https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1577

Permit to disable automatic aliasing when creating a new index.
This is useful when creating a new index that will be switched later (once populate is done for example).


XWB:
:+1:

PR #1762, opened 2021-03-12
Migrate unit / functional tests on github action

deguif:


deguif:
@XWB I think it's ready for now.
I didn't migrate the test coverage part but can be reintroduced later.
This will allow us to unlock releasing ;)

PR #1763, opened 2021-03-12
Test PHP 8.0 with github action

deguif:


deguif:
@XWB PHP 8.0 is now tested successfully in CI.

PR #1771, opened 2021-03-15
Remove BC layer for dynamic templates

deguif:


src/DependencyInjection/Configuration.php
```diff
@@ -91,12 +91,18 @@ private function getDynamicTemplateNode()
...
                                 ->treatNullLike([])
                             ->end()
                         ->end()
                     ->end()
                 ->end()
+                ->validate()
+                    ->ifTrue(static function ($v) {
+                        return 1 !== \count($v);
+                    })
+                    ->thenInvalid('Dynamic template should consist of a single named object: %s.')
```
deguif, line 104: Using `named object` wording as stated in the doc: https://www.elastic.co/guide/en/elasticsearch/reference/7.x/dynamic-templates.html#dynamic-templates

PR #1774, opened 2021-03-15
Fix wrong method call signature in examples

deguif:


doc/usage.md
```diff
@@ -268,12 +268,10 @@ $fieldQuery->setFieldQuery('title', 'I am a title string');
 $fieldQuery->setFieldParam('title', 'analyzer', 'my_analyzer');
 $boolQuery->addShould($fieldQuery);
 
-$tagsQuery = new \Elastica\Query\Terms();
-$tagsQuery->setTerms('tags', ['tag1', 'tag2']);
```
deguif, line 272: `setTerms()` only accepts a single argument now.
This can also be simplified by passing `field` and `terms` directly in constructor.

PR #1776, opened 2021-03-15
Remove import on native php class

deguif:


XWB:
Is it bad practise to import a native php class?

deguif:
@XWB that depends of projects ;)
Here's it's to be consistent with other imports.

PR #1779, opened 2021-03-16
Test with nightly php version

deguif:


XWB:
We have PHP 8.1 covered in the master branch.

PR #1787, opened 2021-04-06
allow Hybrid paginator to use options (like search_type)

bgambier:
I wanted to implement the "dfs_query_then_fetch" option, but HybridPaginator did not have options yet.


XWB:

PR #1798, opened 2021-05-24
Use null coalescing operator

deguif:


src/Manager/RepositoryManager.php
```diff
@@ -69,11 +69,7 @@ public function hasRepository(string $indexName): bool
 
     protected function getRepositoryName(string $indexName): string
     {
-        if (isset($this->indexes[$indexName]['repositoryName'])) {
-            return $this->indexes[$indexName]['repositoryName'];
-        }
-
-        return 'FOS\ElasticaBundle\Repository';
+        return $this->indexes[$indexName]['repositoryName'] ?? 'FOS\ElasticaBundle\Repository';
```
XWB, line 72: How about using `::class constant` ?

deguif: Yes it's done in another PR ;)
Will rebase this one now

deguif: With the rebase, it should be OK now

PR #1820, opened 2021-09-18
Add missing cast to string for object identifiers in ModelToElasticaldentifierTransformer

spiroski:
A while back the casting of the identifier to object was reworked so that it happened only if the object was not a scalar. (PR: #1357)

However when that rework happened this was not implemented in the ModelToElasticaIdentifierTransformer.

This pull requests implements the same logic there as well.


XWB:

PR #1829, opened 2021-10-13
Restore defer config on terminate Listener event after persist - 5.x

### LINKED ISSUE #1756: Doctrine Listener should maintain the original config.defer settings
My application is executing a pub/sub message system.  The consumers can process several messages in the same execution.
I am using the `defer` option with `true` value to persist changes on `kernel.terminate`/`console.terminate` events.

I've seen that `\FOS\ElasticaBundle\Doctrine\Listener` is changing the original configuration in the `onTerminate` method, but it is not restoring the configuration, so it is making some issues in the consumers after the first message is consumed.  The first message is deferred, but the next ones are persisting things during `preFlush`/`postFlush` events.

I am running v5.0.3 but master does have the same thing.
https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/src/Doctrine/Listener.php#L102-L108
```php
    public function onTerminate()
    {
        if ($this->config['defer']) {
            $this->config['defer'] = false;
            $this->persistScheduled();
        }
    }
```

I think the quick solution is to restore the configuration with the original setting.
```php
    public function onTerminate()
    {
        if ($this->config['defer']) {
            $this->config['defer'] = false;
            $this->persistScheduled();
            $this->config['defer'] = true;
        }
    }
```

At the moment, I've created a subscriber with higher priority than the Listener in order to save the original config, from all the listeners, and then after the `onTerminate` event is executed, the saved config is restored.  As `config` is private, I have to use Reflection, thing that I don't really like but I have no choice.

Thanks!

ixarlie:
resolves #1756

### DIFF
--- a/src/Doctrine/Listener.php
+++ b/src/Doctrine/Listener.php
@@ -107,6 +107,7 @@ class Listener
         if ($this->config['defer']) {
             $this->config['defer'] = false;
             $this->persistScheduled();
+            $this->config['defer'] = true;
         }
     }
 


ixarlie:
Hi @XWB

I need some advise to fix the unit/functional tests. My small change shouldn't cause the tests to fail.
Thanks!

PR #1838, opened 2021-11-29
add return type declarations to fixes deprecations in Symfony 5.4

DavidGodefroid:


XWB:
See https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1843

PR #1841, opened 2021-12-09
Don't escape unicode in the query log

b1rdex:
Reasoning: it's very hard to read the log if there are some non-ascii chars in it.


XWB:
Do you have an example?

Test needs to be fixed as well.

b1rdex:
A simple example: query containing some Cyrillic chars. PHP's `json_encode` by default transforms them to Unicode code points. You can see the behavior there: https://3v4l.org/54VOv

b1rdex:
I've rebased the branch and tests pass. Please review.

XWB:
:+1:

PR #1842, opened 2021-12-10
forcemerge takes arguments in query parameters, not in the request body

knusperpixel:
Changed the cookbook entry


XWB:
Thank you.

PR #1844, opened 2021-12-17
Allow Symfony 6

franmomu:
supersedes https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1840

This PR adds a CI job to test with Symfony 6, ~the `ConfigSourcePassTest` has been changed because there was a problem with `replaceArgument` not returning anything and I decided to change it to use an actual `ContainerBuilder`.~ (solved in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1845)

PR #1845, opened 2021-12-17
Remove prophecy

franmomu:
While doing https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1844 I've seen that it was only used in 3 files, so this PR replaces the use of prophecy and also bumps phpunit to 9.5


XWB:
👍

PR #1846, opened 2021-12-22
Return null instead of [] for a null object in Elastica Document

lucasbrtd:
Hello, I don't know if this is a real problem or only my problem.
In fact, looking at the pieces of code below, the first is the current result and the second is the expected result and that I would rather have.
```php
{
  "childOf": [],
  "section": {
    "label": "section"
  },
  "nature": {
    "label": "nature"
  }
}
```
```php
{
  "childOf": null,
  "section": {
    "label": "section"
  },
  "nature": {
    "label": "nature"
  }
}
```
In my fos_elastica.yaml, my "childOf" property is declared as an object.
```yml
app_company:
    properties:
        childOf:
            type: object
            properties:
                id: ~
```

Except that I find that it is not very logical to recover an empty array instead of null for an object whose value is null in database.
What do you think of my proposition ?
Thanks


XWB:
Seems fair to me.

PR #1847, opened 2021-12-29
Allow install psr/log 2 and 3

franmomu:
We implement it in https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/723c6649efadcbf3331daa91e4c5a6e89288d904/src/Logger/ElasticaLogger.php which [is already compatibly with version 2 and 3](https://github.com/php-fig/log/compare/2.0.0...3.0.0).

psr/log 3 cannot be installed because [`elasticsearch-php` doesn't support it yet](https://github.com/elastic/elasticsearch-php/blob/d7539aa7c414a92ee34ae95070da1bc5dd9d30c3/composer.json#L19).

I've also modified the job matrix to add more tests with different versions.

### DIFF
--- a/.github/workflows/continuous-integration.yaml
+++ b/.github/workflows/continuous-integration.yaml
@@ -30,29 +30,34 @@ jobs:
 
   phpunit:
     runs-on: 'ubuntu-20.04'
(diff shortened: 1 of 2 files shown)


XWB:
Thank you :)

PR #1850, opened 2022-01-24
setup per symfony version

sukhoy94:
I suggest to mark that there is no AppKernel.php in newest version of symfony, instead there is bundles.php

### DIFF
--- a/doc/setup.md
+++ b/doc/setup.md
@@ -22,7 +22,7 @@ Instructions for installing and deploying Elasticsearch may be found [here](http
 B: Enable the Bundle
 -------------------------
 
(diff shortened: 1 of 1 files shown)


XWB:
Or just rename the entry to `src/Kernel.php`.

sukhoy94:
@XWB in newest Symfony bundles are included inside config/bundles.php, I think that should be mentioned in docs. Anyway when use flex bundle added automatically

XWB:
That's true. As we don't support Symfony 3 anymore it would be better to remove the entire section.

sukhoy94:
Okay, I've created new PR https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1851

PR #1851, opened 2022-01-24
Update setup.md - enable bundle via adding it to config/bundles.php

sukhoy94:
enable bundle via adding it to config/bundles.php


sukhoy94:
see pls https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1850

XWB:
:+1:  Thank you.

PR #1863, opened 2022-03-31
Fix alias event on AbstractEvent

jcmoro:


XWB:

PR #1881, opened 2022-09-08
Add PHPStan template to HybridResult

bobvandevijver:
This allows users to annotate what transformed object is contained in their HybridResults by annotating their search repositories.

See https://phpstan.org/blog/generics-in-php-using-phpdocs for more information about generics.


XWB:
Thanks :+1:

Issue #1849, opened 2022-01-19
Dynamic Templates fail with: expected array, but got string


arnefi:
Ahoi,

I'm using v6.1.0 and tried to use dynamic templates in my index. According to the [documentation](https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/doc/indexes.md#dynamic-templates) the following definition should be ok:

```
fos_elastica:
    clients:
        default: { url: 'http://elasticsearch:9200' }
    indexes:
        products:
            index_name: products
            persistence:
                driver: orm
                model: "%sylius.model.product_variant.class%"
                provider: ~
                finder: ~
                listener: ~
            dynamic_templates:
                category:
                    match: 'category-*' 
                    mapping:
                        type: keyword
            properties:
                code: ~
```

But when I try to build the index, i get the following error message:

```
# bin/console fos:elastica:populate
Resetting products
 42/42 [============================] 100%
Populating productsRefreshing products

root@d56a25c78558:/srv/app# bin/console fos:elastica:populate

In ArrayNode.php line 260:
                                                                                                                  
  Invalid type for path "fos_elastica.indexes.products.dynamic_templates.category.match". Expected "array", but   
  got "string"
```

I also looked in the `Configuration.php` for the dynamic_templates, but even then my implemention of the yaml looks correct for me:

```
        $node
            ->arrayPrototype()
                ->arrayPrototype()
                    ->children()
                        ->scalarNode('match')->end()
                        ->scalarNode('unmatch')->end()
                        ->scalarNode('match_mapping_type')->end()
                        ->scalarNode('path_match')->end()
                        ->scalarNode('path_unmatch')->end()
                        ->scalarNode('match_pattern')->end()
                        ->arrayNode('mapping')
                            ->variablePrototype()
                                ->treatNullLike([])
                            ->end()
                        ->end()
                    ->end()
                ->end()
                ->validate()
                    ->ifTrue(static function ($v) {
                        return 1 !== \count($v);
                    })
                    ->thenInvalid('Dynamic template should consist of a single named object: %s.')
                ->end()
            ->end()
        ;
```

I tried every combination of Array, Objects, ... but i always end up getting an error. 

The most likely solution according to the error message would be

```
            dynamic_templates:
                category:
                    match: 
                        - 'category-*' 
                    mapping:
                        type: keyword
```

but then I get

```
In ArrayNode.php line 322:
                                                                                                                  
[...]

PR #1884, opened 2022-09-23
[docs]Update Dynamic Templates documentation based on issue

Ferror:
The dynamic_templates expects an array. That's nice, we the example states differently. I was only able to came up to the solution because of the issue below, but no action was done because of that 😢 

https://github.com/FriendsOfSymfony/FOSElasticaBundle/issues/1849

### DIFF
--- a/doc/indexes.md
+++ b/doc/indexes.md
@@ -66,11 +66,11 @@ fos_elastica:
     indexes:
         user:
             dynamic_templates:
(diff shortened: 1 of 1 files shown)


Ferror:
@XWB minor, but important fix :)

XWB (approved):

XWB:
Thanks :)

PR #1892, opened 2022-11-16
Show filtered object count during populate

bobvandevijver:
With the current populate command output you are not given insight in the amount of filtered objects, and they are not used to advance the progress bar as well. Before #1891 this would result in the progress bar not being completed at all, but now it jumps to the end as it is being marked as completed.

I propose the following adjustment: register the amount of filtered objects into the `PreInserObjectsEvent` and forward to the `PostInsertObjectsEvent`. There is also a `PostAsyncInsertObjectsEvent`, but I believe that is not being used at (~~will open a dedicated MR for it~~ #1893). With the data now available to the console logger, we can print the skipped/total object after the populating message, and adjust the progress bar total to show the amount of objects to be really inserted.

I expect test will fail with this change, but before I update them I would love to hear if this is a reasonable change!

### Old output

https://user-images.githubusercontent.com/1835343/202022563-561b2bfe-749f-487a-8b33-cf9af9d5863e.mp4

### New output

https://user-images.githubusercontent.com/1835343/202179821-67ee1dd7-b5b0-4a37-9f28-9cf4cb164470.mp4

### DIFF
--- a/src/Persister/Event/PostInsertObjectsEvent.php
+++ b/src/Persister/Event/PostInsertObjectsEvent.php
@@ -38,15 +38,21 @@ final class PostInsertObjectsEvent extends Event implements PersistEvent
     private $options;
 
     /**
+     * @var int
+     */
+    private $filteredObjectCount;
+
+    /**
      * @param list<object>         $objects
      * @param array<string, mixed> $options
      */
-    public function __construct(PagerInterface $pager, ObjectPersisterInterface $objectPersister, array $objects, array $options)
+    public function __construct(PagerInterface $pager, ObjectPersisterInterface $objectPersister, array $objects, array $options, int $filteredObjectCount = 0)
     {
         $this->pager = $pager;
         $this->objectPersister = $objectPersister;
         $this->objects = $objects;
         $this->options = $options;
+        $this->filteredObjectCount = $filteredObjectCount;
     }
 
     public function getPager(): PagerInterface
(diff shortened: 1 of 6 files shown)


bobvandevijver:
@XWB Anything I can do for this to get merged?

src/Persister/Event/PreInsertObjectsEvent.php
```diff
@@ -107,4 +112,24 @@ public function setObjects($objects)
...
     }
+
+    /**
+     * @param int $count
+     *
+     * @return void
+     *
+     * @internal
+     */
+    public function setFilteredObjectCount($count)
```
XWB, line 123: As we rely on PHP 7.4 it is safe to use type hints and return types.

src/Persister/Event/PreInsertObjectsEvent.php
```diff
@@ -107,4 +112,24 @@ public function setObjects($objects)
...
+    {
+        $this->filteredObjectCount = $count;
+    }
+
+    /**
+     * @return int
+     */
+    public function getFilteredObjectCount()
```
XWB, line 131: You may add a return type.

src/Persister/Event/PostInsertObjectsEvent.php
```diff
@@ -37,16 +37,22 @@ final class PostInsertObjectsEvent extends Event implements PersistEvent
      */
     private $options;
 
+    /**
+     * @var int
+     */
+    private $filteredObjectCount;
```
XWB, line 43:
```suggestion
    private int $filteredObjectCount;
```

XWB (changes requested):

XWB (approved):

XWB:
👍

bobvandevijver:
@XWB I would love to see a release coming soon :)

PR #1893, opened 2022-11-16
Remove non-emitted PostAsyncInsertObjectsEvent

bobvandevijver:
I believe this event is not emitted anywhere from the source, so it should be save to remove.


bobvandevijver:
@XWB Any update on this MR?

XWB:
We don't know if someone uses the event, removing it would be a BC break. Instead, you should deprecate the event so it can be removed in a future major release.

bobvandevijver:
Fair enough, updated!

src/Command/PopulateCommand.php
```diff
@@ -171,13 +170,6 @@ private function populateIndex(OutputInterface $output, string $index, bool $res
             }
         );
 
-        $this->dispatcher->addListener(
```
XWB, line 174: I think this change needs to be reverted, otherwise custom event listeners won't be triggered.

bobvandevijver: I'm not sure what you mean here. This command registers its own listener for an event (that is never emitted at all). This listener registration does not create any event by itself?

XWB (approved):

XWB:
Approved, but there is a merge conflict.

bobvandevijver:
@XWB Merge conflict has been fixed!

XWB:
Thanks :)

PR #1895, opened 2022-12-06
Only require `pagerfanta/core`

mbabker:
Instead of depending on the `pagerfanta/pagerfanta` monopackage (which installs everything), only the `pagerfanta/core` package needs to be a hard dependency.  The subsplit repos for the supported Doctrine integrations are already listed as dev requirements, so this mainly just lightens the production dependencies and allows Composer to work a bit more efficiently (the monorepo has a heavier list of conflicts to deal with because it's trying to set constraints for everything whereas the subsplit packages have rather small dependency lists).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -31,7 +31,7 @@
     "homepage": "https://github.com/FriendsOfSymfony/FOSElasticaBundle",
     "require": {
         "php": "^7.4 || ^8.0",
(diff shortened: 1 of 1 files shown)


jakubtobiasz (approved):

XWB (approved):

XWB:
Seems fair to me.

PR #1897, opened 2022-12-13
Add missing interfaces to TransformedFinder

jakubtobiasz:
Previously, `TransformedFinder`'s public methods were poorly covered with interfaces. This PR adds missing, granular contracts for each public method.

✅ Each public method of `TransformedFinder` is covered by interface
✅ Each interface is granular
✅ Backward-compatible

I wanted to clean up the `PaginatedFinderInterface`, but due to BC reasons I decided to left it as is. Current state doesn't block making this PR working, and we can refactor it later, when new major would be on the horizon.

### DIFF
--- a/src/Finder/HybridFinderInterface.php
+++ b/src/Finder/HybridFinderInterface.php
@@ -1,0 +1,34 @@
+<?php
+
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the FOSElasticaBundle package.
+ *
+ * (c) FriendsOfSymfony <https://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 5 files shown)


jakubtobiasz:
@XWB any chance for review 🤓?

XWB:
Wouldn't this be a BC break? Everyone that extends `TransformedFinder` will have to implement those methods.

jakubtobiasz:
@XWB thanks for your reply! Implementation is already provided in `TransformedFinder` so this change is transparent for the end user. Some methods were already present, but haven't been in any of the existing interfaces.

XWB (approved):

XWB:
Sounds fair to me, thank you for contributing :)

PR #1905, opened 2023-04-06
Remove count from async persister

oleg-andreyev:
Async Persister is returning useless count that is not used anywhere but to calculate it, we are making an expensive COUNT() query.

### DIFF
--- a/tests/Unit/Persister/AsyncPagerPersisterTest.php
+++ b/tests/Unit/Persister/AsyncPagerPersisterTest.php
@@ -13,17 +13,12 @@ namespace FOS\ElasticaBundle\Tests\Unit\Persister;
 
 use FOS\ElasticaBundle\Message\AsyncPersistPage;
 use FOS\ElasticaBundle\Persister\AsyncPagerPersister;
-use FOS\ElasticaBundle\Persister\InPlacePagerPersister;
-use FOS\ElasticaBundle\Persister\ObjectPersisterInterface;
 use FOS\ElasticaBundle\Persister\PagerPersisterInterface;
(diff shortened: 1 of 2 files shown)


tests/Unit/Persister/AsyncPagerPersisterTest.php
```diff
@@ -56,46 +56,4 @@ function ($message) {
...
-            'max_per_page' => 10,
-        ])->willReturn($pagerMock);
-
-        $pagerProviderRegistry = $this->createMock(PagerProviderRegistry::class);
-        $pagerProviderRegistry->expects($this->once())->method('getProvider')->with('foo')->willReturn($provider);
-
-        $messageBus = $this->createMock(MessageBusInterface::class);
-        $sut = new AsyncPagerPersister($pagerPersisterRegistry, $pagerProviderRegistry, $messageBus);
-
-        $sut->insertPage(1, [
```
oleg-andreyev, line 96: funny fact this should have been assert result of `insertPage` but it's no doing it.

oleg-andreyev:
@XWB @toaotc could you please take a look.

XWB:
> Async Persister is returning useless count

Do you know why it is useless? Do you mean the count result is inaccurate?

oleg-andreyev:
> > Async Persister is returning useless count
> 
> Do you know why it is useless? Do you mean the count result is inaccurate?

Based on usage in `AsyncPersistPageHandler`, value is returned and later is not used.
So if value is not used, why do we need to execute it.

XWB (approved):

PR #1906, opened 2023-04-06
added pager_persister to options for AbstractIndexPopulateEvent

oleg-andreyev:
need to have an ability to detect current "persister"


oleg-andreyev:
@XWB @toaotc could you please take a look.

XWB (approved):

PR #1919, opened 2023-08-04
Fix empty dynamic date formats list

7ochem:
If the `dynamic_date_formats` is an empty list
this will produce a `Elastica\Exception\ResponseException` saying:

> class java.util.ArrayList cannot be cast to class java.util.Map (java.util.ArrayList and java.util.Map are in module java.base of loader 'bootstrap')

Screenshot-from-2023-08-04-13-35-54

- friendsofsymfony/elastica-bundle v6.3.1
- ruflin/elastica 7.3.1
- Elasticsearch 6.8.22
- symfony/console v4.4.49

### DIFF
--- a/src/Index/MappingBuilder.php
+++ b/src/Index/MappingBuilder.php
@@ -76,7 +76,7 @@ class MappingBuilder
     {
         $mapping = $indexConfig->getMapping();
 
-        if (null !== $indexConfig->getDynamicDateFormats()) {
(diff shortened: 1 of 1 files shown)


7ochem:
Ok, so this happens with Elasticsearch 6, I need to use Elasticsearch 7 when using friendsofsymfony/elastica-bundle v6... When I use Elasticsearch v7 it's all fine. Closing...

PR #1920, opened 2023-08-04
Client: log query on exception then rethrow it

7ochem:
If an Elastica exception was thrown upon request, then make the request visible that caused it (on verbose console or in your logs), then rethrow the exception.

Example with corrupted request data on the console (calling `php bin/console --env=dev -vvv fos:elastica:populate` very verbose)

### Before:
Screenshot from 2023-08-04 14-02-06
❓ 🤷🏻‍♂️ 

### After:
Screenshot from 2023-08-04 14-04-31
❗ 🕵🏻‍♂️

### DIFF
--- a/src/Elastica/Client.php
+++ b/src/Elastica/Client.php
@@ -13,6 +13,7 @@ namespace FOS\ElasticaBundle\Elastica;
 
 use Elastica\Client as BaseClient;
 use Elastica\Exception\ClientException;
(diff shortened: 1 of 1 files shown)


jasminmistry (approved):

XWB (approved):

XWB:
:+1:

PR #1948, opened 2024-06-06
optimize() deprecated and replaced with forcemerge()

nurikabe:
`optimize()` deprecated in favor of `forcemerge()` in Elastica v5.x and removed in v6.x.


XWB (approved):

XWB:
👍

Issue #1950, opened 2024-07-12
Deprecation of Extension


flavou45:
The "Symfony\Component\HttpKernel\DependencyInjection\Extension" class is considered internal since Symfony 7.1, to be deprecated in 8.1; use Symfony\Component\DependencyInjection\Extension\Extension instead. It may change without further notice. You should not use it from "Lexik\Bundle\JWTAuthenticationBundle\DependencyInjection\LexikJWTAuthenticationExtension".

XWB:
Fixed in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1959

PR #1959, opened 2025-02-13
Use correct Extension class per Symfony deprecation

Lustmored:
As per #1950 the `Symfony\Component\HttpKernel\DependencyInjection\Extension` is deprecated and `Symfony\Component\DependencyInjection\Extension\Extension` should be used.

This class is available in all Symfony versions supported by this bundle.


XWB (approved):

XWB:
Thank you.

Issue #1618, opened 2020-04-16
How to define multiple template patterns for an index template?


benjamintoussaint:
Hi,

I've following this doc to create an index template:
https://github.com/FriendsOfSymfony/FOSElasticaBundle/blob/master/doc/templates.md#index-templates

It gives `template: some_index_*`  as an example pattern and refers to the official documentation:
https://www.elastic.co/guide/en/elasticsearch/reference/6.6/indices-templates.html#indices-templates

But in the official documentation, I see that we can define an array of patterns:
`"index_patterns": ["te*", "bar*"],`

Is it possible to apply a template using multiple patterns and then how?
If not, how could I manually apply a given template to multiple indexes?

Thank you!

XWB:
Should be fixed in https://github.com/FriendsOfSymfony/FOSElasticaBundle/pull/1962

PR #1962, opened 2025-02-28
Updated index template feature to be compatible with ES 6.x+

### LINKED ISSUE #1961: Index Template configuration has not been updated from Elasticsearch 5.6
I'm working on a PR to address this, so I'm opening this issue primarily to document that the problem was found and to have an issue# to reference in the PR.

In ES 5.6 and earlier, the configuration for index templates ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/indices-templates.html)) used a `template` property to define a single index pattern for which to apply the template

```
PUT _template/template_1
{
  "template": "te*",
  ...
```

This is the configuration that FOSElasticaBundle generates.

This was changed in 6.0 ([docs](https://www.elastic.co/guide/en/elasticsearch/reference/6.0/indices-templates.html)) to an `index_patterns` property that accepts an array of patterns.

```
PUT _template/template_1
{
  "index_patterns": ["te*", "bar*"],
   ...
```

FOSElasticaBundle does not appear to have been updated to generate this new configuration key.  Thus, when trying to reset the templates, it fails:

```
Resetting all templates

In Http.php line 178:
                                          
  unknown key [template] in the template  
                                          

fos:elastica:reset-templates [--index INDEX] [--force-delete]
```

The missing array support was noted in #1618 — but unless I'm missing something; please call me out if I am — as far as I can see no one has flagged that templates no longer seem to work *at all*.

As I said I'm working on a fix for this, and I'll reference this and the other issue in the PR.

mwoynarski:
Fixes #1961, #1618

This PR updates the configuration of index templates to be compatible with Elasticsearch 6.0 and up.

Compatibility is maintained with the old configuration format.


XWB (approved):

XWB:
👍

PR #1964, opened 2025-03-13
Add events for requests

### LINKED ISSUE #1963: [Feature] Would you consider a Request event?
Hello.

I wanted to add transactions for sentry to track all requests being passed to ES through the bundle, but at this time it seems like my only option is to decorate the Client class.
I could work on a PR which would add an event, which would fire each time after a request was successfully (or not) made, along with a PreRequest event.

It could be controlled by not passing the `event_distributor`  class, and would only be needed in the bundle itself, as obviously the elastica client doesn't use that. Or it could be done in the library below - either way works.

Please let me know if you'd consider allowing such a feature.

It would boil down to 3 events (`PreRequest`, `PostRequest`, `RequestFailure`), + nullable dispatch call to event distributor.

pkly:
Fixes #1963 

Things to consider: at this time it's not possible to pass the full request object in the `PreRequest` event as that object is created and used by the client below. It could be done if the library would change it so that it's in a separate method so we could override it.

The settting of `EventDispatcher` could also just be done in the constructor at this point.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,8 @@
         "symfony/dependency-injection": "^5.4 || ^6.4 || ^7.1",
         "symfony/framework-bundle": "^5.4 || ^6.4 || ^7.1",
         "symfony/property-access": "^5.4 || ^6.4 || ^7.1",
-        "symfony/stopwatch": "^5.4 || ^6.4 || ^7.1"
+        "symfony/stopwatch": "^5.4 || ^6.4 || ^7.1",
+        "symfony/event-dispatcher": "^5.4 || ^6.4 || ^7.1"
     },
     "require-dev": {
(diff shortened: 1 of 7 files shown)


XWB (approved):

XWB:
> Things to consider: at this time it's not possible to pass the full request object in the PreRequest event as that object is created and used by the client below. It could be done if the library would change it so that it's in a separate method so we could override it.

I think that's fine for now. Does the `PreRequest` provides all the information you need for now?

pkly:
Yes, I already decorated the `request` method in our application and it works essentially the same, so this should be fine.

XWB:
Thanks for you hard work 👍

Issue #1968, opened 2025-04-18
Any way of disabling update when no indexed fields are changed


magnetik:
Hello,

Let's say I'm indexing in ES the fields `A` and `B` in an entity that contains `A`, `B`, `C` and `D`.

When I'm updating the field `C` of my entity, an ES update is triggered while it's not necessary.

Have you every considered trying to detect which fields needs to be updated?

PR #1969, opened 2025-05-14
Implement ConditionalUpdate interface for selective entity updates

magnetik:
It's a proposition for #1968.

Some tests are red but I think it's related to a dependency change.

### DIFF
--- a/src/Doctrine/ConditionalUpdate.php
+++ b/src/Doctrine/ConditionalUpdate.php
@@ -1,0 +1,20 @@
+<?php
+
+/*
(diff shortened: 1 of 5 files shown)


XWB:
Seems fine to me, but many tests are failing.

magnetik:
The failing tests are not related to this PR. Fixed in #1970

XWB (approved):

XWB:
:+1:

magnetik:
Thanks for the fast merge ! Do you think a release will be made soon?

Is there any work needs to be done before?

PR #1986, opened 2026-03-20
Symfony 8.0 Support

bojanpejic:
- most significant changes are changing configuration files from XML to PHP
- added return type declerations
- Symfony 8.0 requires at least PHP version 8.4, reflecting that in continuous-integrataion.yaml
- updates regarding the php-cs-fixer configuration to exclude certain methods


trsteel88:
@XWB is this going to be tagged as a version anytime soon?

XWB:
Done

